PR #13, opened 2017-04-06
Select Audio Track

Mecharyry:
#### Problem
Clients of `no-player` are unable to select the `audio track` they desire at any point during video playback. 

#### Solution
Expose methods in the `Player` to allow clients to get a list of `PlayerAudioTrack`. `PlayAudioTrack` consists of basic info around a `MediaFormat` such as `language`, `channels`, `mimeType` etc. To set an `AudioTrack` the client needs to specify an index from the `PlayerAudioTrack` that matches the audio track they desire. 

`mediaPlayer.selectTrack()` is only available above `api` level 15 so we upped this as well in this PR. It seems impossible to set an `audio track` before this.

#### Paired with
@jackSzm

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/Player.java
+++ b/core/src/main/java/com/novoda/noplayer/Player.java
@@ -2,7 +2,6 @@ package com.novoda.noplayer;
 
 import android.net.Uri;
 
-import com.google.android.exoplayer.MediaFormat;
 import com.novoda.noplayer.exoplayer.Bitrate;
 import com.novoda.noplayer.player.PlayerInformation;
 
@@ -32,7 +31,7 @@ public interface Player extends PlayerState, PlayerListeners {
 
     void setAudioTrack(int mediaTrackPosition);
 
-    List<MediaFormat> getAudioTracks();
+    List<PlayerAudioTrack> getAudioTracks();
 
     interface PreReleaseListener {
 
--- a/core/src/main/java/com/novoda/noplayer/Player.java
+++ b/core/src/main/java/com/novoda/noplayer/Player.java
@@ -29,7 +29,7 @@ public interface Player extends PlayerState, PlayerListeners {
 
     void attach(PlayerView playerView);
 
-    void setAudioTrack(int mediaTrackPosition);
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/novoda/noplayer/Player.java
```diff
@@ -27,6 +29,10 @@
 
     void attach(PlayerView playerView);
 
+    void setAudioTrack(int audioTrackIndex);
```
ouchadam: this is more of a `selectAudioTrack(index)` than a set

ouchadam:
nice work! just the comment about a method name, otherwise LGMT 👍

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -236,4 +238,35 @@ private boolean isInPlaybackState() {
...
+        ArrayList<PlayerAudioTrack> audioTracks = new ArrayList<>();
+        for (MediaPlayer.TrackInfo trackInfo : mediaPlayer.getTrackInfo()) {
+            if (trackInfo.getTrackType() == MediaPlayer.TrackInfo.MEDIA_TRACK_TYPE_AUDIO) {
+                audioTracks.add(new PlayerAudioTrack(String.valueOf(trackInfo.hashCode()), trackInfo.getLanguage(), "", -1, -1));
+            }
+        }
+        return audioTracks;
+    }
+
+    public void setAudioTrack(int audioTrackIndex) {
```
juankysoriano: same as the other? `select`

juankysoriano:
Same as @ouchadam ; `select` seems a better name than `set`

jackSzm:
Amended in : https://github.com/novoda/no-player/pull/13/commits/3fef8ed552a4262553ebd3763ade9f16283f8847 

Also found one more thing 🙈 . https://github.com/novoda/no-player/pull/13/commits/3fef8ed552a4262553ebd3763ade9f16283f8847#diff-379a1e7be0f368e89f56122346e736a4R443

PR #18, opened 2017-05-24
ExoPlayer2/ Listeners and forwarders

Mecharyry:
## Problem
The architecture behind `no-player` goes something like this.

The client requests a `Player` which can be either an `ExoPlayer` or an `AndroidMediaPlayer`, each of which has an underlying `facade`(forgive the naming) which talks directly to either the real `exoplayer` or `mediaplayer`. `NoPlayer` exposes implementation agnostic listeners that the clients can use. The `facades` have implementation specific listeners and forwarders (named listeners) that transform the implementation specific events into no-player events and forward to the relevant no-player listeners.

The problem is, that nearly everything is called a listener :laughing: which makes working on `no-player` confusing

Furthermore, because there are different events, some of the them have different interfaces and some of them even implement some interfaces and no-op them and then extend again only using the ones they need. Confusing!

## Solution
Have one `Listener` that encapsulates all of the events that need to be mapped between `exo-player` and `no-player`. Implement a `no-op` version of `Listener` and then create additional forwarders from this that implement the methods they need to know about that map to `no-player`.

## Paired with
@jackSzm and @Dorvaryn


Mecharyry:
Something is up with the MediaPlayer 😢

build.gradle
```diff
@@ -1,18 +1,18 @@
...
 
 buildscript {
     repositories {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.3'
-        classpath 'com.novoda:bintray-release:0.4.0'
+        classpath 'com.android.tools.build:gradle:2.3.1'
+        classpath 'com.novoda:bintray-release:0.5.0'
```
Mecharyry: Updated all these dependencies so that we can have a composite build in another project, to avoid shipping this halfway house code 😄

core/src/main/java/com/novoda/noplayer/NoPlayerView.java
```diff
@@ -86,7 +93,24 @@ public void removeControls() {
...
         public void onVideoSizeChanged(int width, int height, int unappliedRotationDegrees, float pixelWidthHeightRatio) {
-            aspectRatioChangeListener.onVideoSizeChanged(width, height, unappliedRotationDegrees, pixelWidthHeightRatio);
+            aspectRatioChangeListener.forwardVideoSizeChanged(width, height, unappliedRotationDegrees, pixelWidthHeightRatio);
+        }
+    };
+
+    private final Player.StateChangedListener stateChangedListener = new Player.StateChangedListener() {
+        @Override
+        public void onVideoPlaying() {
+            shutterView.setVisibility(INVISIBLE);
```
Mecharyry: We opted to use the `SimpleExoPlayerView` for both `MediaPlayer` and `ExoPlayer` but we found that the `shutterView` is rendered on top of the `surfaceView` to prevent artifacts appearing during the prepare phase of setting up the video. Due to this being used on `MediaPlayer` we never reach the `onFirstFrameRendered` call which would have hidden the shutter, so now we grab the shutter ourselves and hide it when the playback starts.

core/src/main/java/com/novoda/noplayer/NoPlayerView.java
```diff
@@ -29,15 +30,16 @@ public NoPlayerView(Context context, AttributeSet attrs) {
     public NoPlayerView(Context context, AttributeSet attrs, int defStyleAttr) {
         super(context, attrs, defStyleAttr);
         surfaceHolderProvider = new PlayerViewSurfaceHolder();
-        aspectRatioChangeListener = new AspectRatioChangeListener(this);
+        aspectRatioChangeListener = new AspectRatioChangeForwarder(this);
```
juankysoriano: `aspectRatioChangeForwarder`

core/src/main/java/com/novoda/noplayer/NoPlayerView.java
```diff
@@ -29,15 +30,16 @@ public NoPlayerView(Context context, AttributeSet attrs) {
...
     }
 
     @Override
     protected void onFinishInflate() {
         super.onFinishInflate();
         View.inflate(getContext(), R.layout.noplayer_view, this);
-        videoFrame = Views.findById(this, R.id.video_frame);
+        videoFrame = Views.findById(this, R.id.exo_content_frame);
```
juankysoriano: This the `NoPlayerView` however we are using a layout named `exo_content_frame`. Is that ok?

Mecharyry: This is internal to `SimpleExoPlayerView` which is a `FrameLayout` that performs some operations on an `ExoPlayer` however, this is a `FrameLayout` whose main purpose is to expose things like a "Controller", "SubtitlesView" and "SurfaceView" (for rendering video). We reasoned that we could reuse this view for both the `MediaPlayer` and `ExoPlayer` because in terms of the `MediaPlayer` we only care for the internal `SurfaceView` and all other methods are ignored.

Mecharyry: We could, and I am not sure how feasible this is, just use a plain `SurfaceView` for both `MediaPlayer` and `ExoPlayer`. This would likely be a lot of work for `ExoPlayer2` because the `SimpleExoPlayerView` does a lot of setup for rendering internally.

core/src/main/java/com/novoda/noplayer/exoplayer/EventListener.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.android.exoplayer2.ExoPlaybackException;
+import com.google.android.exoplayer2.ExoPlayer;
+import com.google.android.exoplayer2.PlaybackParameters;
+import com.google.android.exoplayer2.Timeline;
+import com.google.android.exoplayer2.source.TrackGroupArray;
+import com.google.android.exoplayer2.trackselection.TrackSelectionArray;
+
+import java.util.List;
+
+class EventListener implements ExoPlayer.EventListener {
```
juankysoriano: Is this a listener or a forwarder?

juankysoriano: I ask because it is a listener but seems to be composed by forwarders.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerExtractorMediaSourceListener.java
```diff
@@ -0,0 +1,22 @@
+package com.novoda.noplayer.exoplayer;
+
+import com.google.android.exoplayer2.source.ExtractorMediaSource;
+
+import java.io.IOException;
+import java.util.List;
+
+class ExoPlayerExtractorMediaSourceListener implements ExtractorMediaSource.EventListener {
```
juankysoriano: Same question as before

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerEventForwarder.java
```diff
@@ -0,0 +1,98 @@
...
+import android.view.Surface;
+
+import com.google.android.exoplayer2.ExoPlaybackException;
+import com.google.android.exoplayer2.Format;
+import com.google.android.exoplayer2.decoder.DecoderCounters;
+import com.google.android.exoplayer2.upstream.DataSpec;
+
+import java.io.IOException;
+
+abstract class ExoPlayerEventForwarder implements Forwarder {
```
Mecharyry: All `Forwarder`s were collapsed into one interface to make it easier to understand and forward events between the `ExoPlayerTwoFacade` and the `ExoPlayerTwoImpl`. There are children forwarders of this that know only the methods they are interested in that perform the conversion from `ExoPlayer` to `NoPlayer` callbacks.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoFacade.java
```diff
@@ -4,64 +4,70 @@
...
-
     private final SimpleExoPlayer exoPlayer;
     private final MediaSourceFactory mediaSourceFactory;
-
-    private List<Listener> listeners = new CopyOnWriteArrayList<>();
-    private InfoListeners infoListeners;
-    private BitrateChangedListeners bitrateChangedListeners;
-    private InternalErrorListener internalErrorListener;
+    private final EventListener exoPlayerEventListener;
+    private final MediaSourceEventListener mediaSourceEventListener;
```
Mecharyry: Extract all of the different "Event" Listeners in `ExoPlayer` into separate classes passing each a reference to `List<Forwarders>`, personally I don't like that we can make state changes on `List<Forwarders>` here and it will affect all of the other listeners but we couldn't come up with a nice way of doing this because of the creation order.

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -17,27 +17,28 @@
...
         Log.setShowLogs(true);
         setContentView(R.layout.activity_main);
+        playerView = (PlayerView) findViewById(R.id.player_view);
+    }
 
-        player = new PlayerFactory(this, PrioritisedPlayers.prioritiseExoPlayer()).create();
+    @Override
+    protected void onStart() {
+        super.onStart();
+        player = new PlayerFactory(this, PrioritisedPlayers.prioritiseMediaPlayer()).create();
```
juankysoriano: I like the change but I wonder if it is safe to do the change. Given my lack of knowledge on this area...

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -139,7 +138,6 @@ public PlayerInformation getPlayerInformation() {
     public void attach(PlayerView playerView) {
         videoContainer = VideoContainer.with(playerView.getContainerView());
         facade.setPlayer(playerView.simplePlayerView());
-        addVideoSizeChangedListener(playerView.getVideoSizeChangedListener());
```
Mecharyry: This is now done internally in the `SimpleExoPlayerView`

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -158,7 +156,4 @@ private void showContainer() {
         videoContainer.show();
     }
 
-    public ExoPlayerTwoFacade getInternalExoPlayer() {
```
Mecharyry: This made me 😢 

We have `Player` and `ExoPlayerTwoImpl` to prevent clients from having to deal with the underlying `ExoPlayer` but this method provides just that.

core/src/main/java/com/novoda/noplayer/exoplayer/InfoForwarder.java
```diff
@@ -0,0 +1,221 @@
...
+    private final InfoListeners infoListeners;
+
+    InfoForwarder(InfoListeners infoListeners) {
+        this.infoListeners = infoListeners;
+    }
+
+    @Override
+    public void onLoadStarted(DataSpec dataSpec, int dataType, int trackType, Format trackFormat, int trackSelectionReason,
+                              Object trackSelectionData, long mediaStartTimeMs, long mediaEndTimeMs, long elapsedRealtimeMs) {
+        HashMap<String, String> keyValuePairs = new HashMap<>();
```
Mecharyry: All of these methods were originally exposed to `NoPlayer` when they should not be. These are specific to `ExoPlayer` and we should be providing a `NoPlayer` version, this is that 😄

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -229,21 +233,23 @@ public boolean isPlaying() {
...
-        return mediaPlayer != null
+        return hasPlayer()
                 && currentState != STATE_ERROR
                 && currentState != STATE_IDLE
                 && currentState != STATE_PREPARING;
     }
 
     public void stop() {
-        mediaPlayer.stop();
+        if (hasPlayer()) {
```
Mecharyry: Yea... 😄

core/src/main/java/com/novoda/noplayer/exoplayer/InfoListener.java
```diff
@@ -1,30 +1,12 @@
...
-
-import java.io.IOException;
+import java.util.Map;
 
 /**
  * A listener for debugging information.
  */
 public interface InfoListener {
 
-    void onDroppedFrames(int count, long elapsed);
```
juankysoriano: What's the reason behind this change on this interface? Didnt' see it in the diff

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -318,6 +325,12 @@ public void release() {
         getPlayerReleaseListener().onPlayerPreRelease(this);
         mediaPlayer.release();
         getStateChangedListeners().onVideoReleased();
+        removeVideoSizeChangedListener(videoSizeChangedListener);
+        removeStateChangedListener(stateChangedListener);
+        removeHeartbeatCallback(heartbeatCallback);
+        videoSizeChangedListener = null;
```
Mecharyry: Adding Listeners? Should probably remove them too 😄

core/src/main/java/com/novoda/noplayer/exoplayer/InfoListener.java
juankysoriano: Nevermind, I have just seen the `InfoForwarder`

core/src/main/java/com/novoda/noplayer/exoplayer/EventListener.java
Mecharyry: This is a bit funny actually 😄  so this listens to `exoPlayer` events and passes them to the forwarders. I guess we could call it a forwarder too but it's more important that it listens to events... I guess 🤷‍♂️

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerExtractorMediaSourceListener.java
Mecharyry: Same answer as before 👅

core/src/main/java/com/novoda/noplayer/exoplayer/EventListener.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private final List<Forwarder> forwarders;
+
+    EventListener(List<Forwarder> forwarders) {
+        this.forwarders = forwarders;
+    }
+
+    @Override
+    public void onTimelineChanged(Timeline timeline, Object manifest) {
+        // no-op
```
zegnus: Why we are not forwarding these ones?

demo/src/main/java/com/novoda/demo/MainActivity.java
Mecharyry: This is perfectly safe and really we should recreate the `Player` everytime we release it. This might require some rework in other projects depending on which method calls they use but for the most part it should be fine. Hint: don't call release when you want to use the Player again 😄

core/src/main/java/com/novoda/noplayer/exoplayer/EventListener.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.android.exoplayer2.trackselection.TrackSelectionArray;
+
+import java.util.List;
+
+class EventListener implements ExoPlayer.EventListener {
+
+    private final List<Forwarder> forwarders;
+
+    EventListener(List<Forwarder> forwarders) {
+        this.forwarders = forwarders;
```
zegnus: `this.forwarders = new ArrayList<>(forwarders)` otherwise this is a pointer and there is no guarantee that someone else will remove an element while you're looping ;)

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerExtractorMediaSourceListener.java
```diff
@@ -0,0 +1,22 @@
...
+
+    private final List<Forwarder> forwarders;
+
+    ExoPlayerExtractorMediaSourceListener(List<Forwarder> forwarders) {
+        this.forwarders = forwarders;
```
zegnus: same, create a copy

core/src/main/java/com/novoda/noplayer/exoplayer/InfoListener.java
Mecharyry: The reasoning behind this is:

The forwarders have knowledge of exo-player, the listeners are the ones exposed through `no-player` and are not implementation specific. This allows the client to sign up to one listener that can be used for both implementations.

The methods here now appear in one of the exo-player listeners and get transformed in the `InfoForwarder`.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoFacade.java
```diff
@@ -4,64 +4,70 @@
...
         DefaultDataSourceFactory defaultDataSourceFactory = new DefaultDataSourceFactory(context, "user-agent");
         Handler handler = new Handler(Looper.getMainLooper());
         DefaultTrackSelector trackSelector = new DefaultTrackSelector();
         SimpleExoPlayer exoPlayer = ExoPlayerFactory.newSimpleInstance(context, trackSelector, new DefaultLoadControl());
         MediaSourceFactory mediaSourceFactory = new MediaSourceFactory(defaultDataSourceFactory, handler);
-        return new ExoPlayerTwoFacade(exoPlayer, mediaSourceFactory);
-    }
 
-    public ExoPlayerTwoFacade(SimpleExoPlayer exoPlayer, MediaSourceFactory mediaSourceFactory) {
+        List<Forwarder> forwarders = new CopyOnWriteArrayList<>();
```
zegnus: Can wrap in a first class collection

core/src/main/java/com/novoda/noplayer/exoplayer/EventListener.java
Mecharyry: Because we do not know what they map to. This needs to be defined in `no-player` before it can be given to a forwarder.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -30,14 +30,13 @@ public ExoPlayerTwoImpl(ExoPlayerTwoFacade facade) {
...
         prepareForwarder = new OnPrepareForwarder(getPreparedListeners(), this);
-        facade.addListener(prepareForwarder);
-        facade.addListener(new OnCompletionForwarder(getCompletionListeners()));
-        facade.addListener(new OnErrorForwarder(this, getErrorListeners()));
-        facade.addListener(new BufferStateForwarder(getBufferStateListeners()));
-        facade.addListener(new VideoSizeChangedForwarder(getVideoSizeChangedListeners()));
-
-        facade.setBitrateChangedListeners(getBitrateChangedListeners());
-        facade.setInfoListeners(getInfoListeners());
+        facade.addForwarder(prepareForwarder);
```
zegnus: for convenience you could create a `... addForwarder(Forwarder... forwarder)` method

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerVideoRendererEventListener.java
```diff
@@ -0,0 +1,70 @@
...
+
+class ExoPlayerVideoRendererEventListener implements VideoRendererEventListener {
+
+    private final List<Forwarder> forwarders;
+
+    private int videoWidth;
+    private int videoHeight;
+
+    ExoPlayerVideoRendererEventListener(List<Forwarder> forwarders) {
+        this.forwarders = forwarders;
```
zegnus: copy

core/src/main/java/com/novoda/noplayer/exoplayer/InfoForwarder.java
```diff
@@ -0,0 +1,221 @@
...
+class InfoForwarder extends ExoPlayerEventForwarder {
+
+    private final InfoListeners infoListeners;
+
+    InfoForwarder(InfoListeners infoListeners) {
+        this.infoListeners = infoListeners;
+    }
+
+    @Override
+    public void onLoadStarted(DataSpec dataSpec, int dataType, int trackType, Format trackFormat, int trackSelectionReason,
```
zegnus: If you extract all these params in a class, then you can add logic into that object in order to return the hashmap (also you can test in isolation this new class) something like `exoPlayerLoadEvent.toMap()`

core/src/main/java/com/novoda/noplayer/exoplayer/MediaSourceEventListener.java
```diff
@@ -0,0 +1,153 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+class MediaSourceEventListener implements AdaptiveMediaSourceEventListener {
+
+    private final List<Forwarder> forwarders;
+
+    MediaSourceEventListener(List<Forwarder> forwarders) {
+        this.forwarders = forwarders;
```
zegnus: copy

core/src/main/java/com/novoda/noplayer/exoplayer/MediaSourceEventListener.java
```diff
@@ -0,0 +1,153 @@
...
+                                int trackType,
+                                Format trackFormat,
+                                int trackSelectionReason,
+                                Object trackSelectionData,
+                                long mediaStartTimeMs,
+                                long mediaEndTimeMs,
+                                long elapsedRealtimeMs,
+                                long loadDurationMs,
+                                long bytesLoaded) {
+
```
zegnus: extra line not needed

core/src/main/java/com/novoda/noplayer/exoplayer/MediaSourceEventListener.java
```diff
@@ -0,0 +1,153 @@
...
+                    trackSelectionReason,
+                    trackSelectionData,
+                    mediaStartTimeMs,
+                    mediaEndTimeMs,
+                    elapsedRealtimeMs,
+                    loadDurationMs,
+                    bytesLoaded
+            );
+        }
+
```
zegnus: extra line non needed

core/src/main/java/com/novoda/noplayer/exoplayer/InfoForwarder.java
juankysoriano: I agree with @zegnus , it seems that every callback has almost the same parameters so extracting to a class should be easy

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoFacade.java
```diff
@@ -4,75 +4,81 @@
...
 import com.novoda.noplayer.ContentType;
-import com.novoda.noplayer.listeners.BitrateChangedListeners;
-import com.novoda.noplayer.listeners.InfoListeners;
 
-import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
-public class ExoPlayerTwoFacade implements VideoRendererEventListener {
+class ExoPlayerTwoFacade {
```
Mecharyry: This class is what communicates with `ExoPlayer` it should not be accessible outside of the library, all clients should go through the `Player` interface and its implementations.

Mecharyry:
`ExoPlayerForwarder` now implements all event listeners that we are interested in `ExoPlayer` and we forward all events. There is an `ExoPlayerEventForwarder` which is an abstract class that swallows all events by default. We add specific `ExoPlayerEventForwarders` based on the `Player` listeners that we want to map to. The forwarders perform the mapping.

core/src/main/java/com/novoda/noplayer/NoPlayerView.java
juankysoriano: I think we can leave it like this

PR #21, opened 2017-05-26
Refactor exo player two facade

Mecharyry:
⚠️ Clone of #18 as it was accidentally recreated 😬  ⚠️ 

## Problem
The architecture behind `no-player` goes something like this.

The client requests a `Player` which can be either an `ExoPlayer` or an `AndroidMediaPlayer`, each of which has an underlying `facade`(forgive the naming) which talks directly to either the real `exoplayer` or `mediaplayer`. `NoPlayer` exposes implementation agnostic listeners that the clients can use. The `facades` have implementation specific listeners and forwarders (named listeners) that transform the implementation specific events into no-player events and forward to the relevant no-player listeners.

The problem is, that nearly everything is called a listener :laughing: which makes working on `no-player` confusing

Furthermore, because there are different events, some of the them have different interfaces and some of them even implement some interfaces and no-op them and then extend again only using the ones they need. Confusing!

## Solution
Have one `Listener` that encapsulates all of the events that need to be mapped between `exo-player` and `no-player`. Implement a `no-op` version of `Listener` and then create additional forwarders from this that implement the methods they need to know about that map to `no-player`.

## Paired with
@jackSzm and @Dorvaryn


core/src/main/java/com/novoda/noplayer/exoplayer/InfoForwarder.java
```diff
@@ -0,0 +1,210 @@
...
+class InfoForwarder extends ExoPlayerEventForwarder {
+
+    private final InfoListeners infoListeners;
+
+    InfoForwarder(InfoListeners infoListeners) {
+        this.infoListeners = infoListeners;
+    }
+
+    @Override
+    public void onLoadStarted(DataSpec dataSpec, int dataType, int trackType, Format trackFormat, int trackSelectionReason,
```
zegnus: Vertical aligment

Mecharyry: Will make sure that this is changed in the next PR.

PR #23, opened 2017-05-30
Exo-player two facade and forwarders

Mecharyry:
## Problem
`ExoPlayerForwarder` contained all of the events that we listen to from `ExoPlayer` and we created `Forwarders` that adapted the `ExoPlayer` events to our `no-player` events. Some of these `Forwarders` contained state. `Facade` and `Imple` were mixing responsibilities with no clear boundary.

## Solution
- Remove the catch all `Forwarder` and implement specific forwarders for the events we want to listen to.
- Create a generic forwarder that has as collaborators the specific forwarder implementations.
- Remove the `Facade` from exo-player because the code is small enough to be encapsulated in one class with collaborators.

### Test(s) added 
No tests atm, we hope to branch off and have these covered soon.

### Screenshots
No UI changes.

### Paired with 
@Dorvaryn and @jackSzm

PR #24, opened 2017-05-30
Refactoring media player listeners

Mecharyry:
## Problem
- `ExoPlayer` and `MediaPlayer` handle listeners in different ways.
- The `MediaPlayerFacade` is quite complex and is exposed as a public class.

## Solution
As per #23 we want to have the same forwarding logic in `MediaPlayer` to enforce consistency across both parts of `no-player`.

-  Reduce the scope of `MediaPlayerFacade` to package-protected, it shouldn't be used by clients. 

### Test(s) added 
No tests yet, we want to start adding these soon!

### Screenshots
No UI changes.

### Paired with 
@Dorvaryn and @jackSzm

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/mediaplayer/CheckBufferHeartbeatCallback.java
+++ b/core/src/main/java/com/novoda/noplayer/mediaplayer/CheckBufferHeartbeatCallback.java
@@ -6,7 +6,7 @@ import com.novoda.noplayer.VideoPosition;
 
 @SuppressWarnings("PMD.RedundantFieldInitializer")
         // we're being very explicit with our default field values, not a bad thing!
-class CheckBufferHeartbeatCallback implements Heart.Heartbeat.Callback<Player> {
+public class CheckBufferHeartbeatCallback implements Heart.Heartbeat.Callback<Player> {
 
     private static final int FORCED_BUFFERING_BEATS_THRESHOLD = 4;
     private final BufferListener bufferListener;
@@ -14,7 +14,7 @@ class CheckBufferHeartbeatCallback implements Heart.Heartbeat.Callback<Player> {
     private VideoPosition previousPosition = VideoPosition.INVALID;
     private int beatsPlayed;
 
-    CheckBufferHeartbeatCallback(BufferListener bufferListener) {
+    public CheckBufferHeartbeatCallback(BufferListener bufferListener) {
         this.bufferListener = bufferListener;
     }
 
--- a/core/src/main/java/com/novoda/noplayer/mediaplayer/CheckBufferHeartbeatCallback.java
+++ b/core/src/main/java/com/novoda/noplayer/mediaplayer/CheckBufferHeartbeatCallback.java
@@ -4,8 +4,6 @@ import com.novoda.noplayer.Heart;
 import com.novoda.noplayer.Player;
 import com.novoda.noplayer.VideoPosition;
 
-@SuppressWarnings("PMD.RedundantFieldInitializer")
-        // we're being very explicit with our default field values, not a bad thing!
 public class CheckBufferHeartbeatCallback implements Heart.Heartbeat.Callback<Player> {
 
     private static final int FORCED_BUFFERING_BEATS_THRESHOLD = 4;
@@ -14,7 +12,7 @@ public class CheckBufferHeartbeatCallback implements Heart.Heartbeat.Callback<Pl
     private VideoPosition previousPosition = VideoPosition.INVALID;
     private int beatsPlayed;
 
-    public CheckBufferHeartbeatCallback(BufferListener bufferListener) {
+    CheckBufferHeartbeatCallback(BufferListener bufferListener) {
         this.bufferListener = bufferListener;
     }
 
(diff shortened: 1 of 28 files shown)


core/src/main/java/com/novoda/noplayer/mediaplayer/CheckBufferHeartbeatCallback.java
```diff
@@ -6,15 +6,15 @@
 
 @SuppressWarnings("PMD.RedundantFieldInitializer")
```
ouchadam: this warning probably isn't needed anymore since I don't see any defaults being initialised

core/src/main/java/com/novoda/noplayer/mediaplayer/forwarder/CompletionInfoForwarder.java
```diff
@@ -0,0 +1,24 @@
...
+    private final InfoListeners infoListeners;
+
+    public CompletionInfoForwarder(InfoListeners infoListeners) {
+        this.infoListeners = infoListeners;
+    }
+
+    @Override
+    public void onCompletion(MediaPlayer mp) {
+        HashMap<String, String> keyValuePairs = new HashMap<>();
+        keyValuePairs.put("mp", String.valueOf(mp));
```
ouchadam: `String.valueOf(mp)`? what is this doing?

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -30,7 +30,7 @@ protected void onCreate(Bundle savedInstanceState) {
     @Override
     protected void onStart() {
         super.onStart();
-        player = new PlayerFactory(this, PrioritisedPlayers.prioritiseExoPlayer()).create();
+        player = new PlayerFactory(this, PrioritisedPlayers.prioritiseMediaPlayer()).create();
```
Mecharyry: Reverted this. We added a comment at the tail end of the last PR to add a switch at some point in the UI.

core/src/main/java/com/novoda/noplayer/mediaplayer/forwarder/CompletionInfoForwarder.java
Mecharyry: Could change this but then it will stop working for the primitives that also come through on the interfaces. I would rather leave as a `String` because at the end of the day it is expected to be used for logging.

Mecharyry: Well, all of the info listeners are used for logging really. Client's shouldn't really rely on it for anything else.

core/src/main/java/com/novoda/noplayer/mediaplayer/CheckBufferHeartbeatCallback.java
```diff
@@ -4,9 +4,7 @@
 import com.novoda.noplayer.Player;
 import com.novoda.noplayer.VideoPosition;
 
-@SuppressWarnings("PMD.RedundantFieldInitializer")
-        // we're being very explicit with our default field values, not a bad thing!
-class CheckBufferHeartbeatCallback implements Heart.Heartbeat.Callback<Player> {
+public class CheckBufferHeartbeatCallback implements Heart.Heartbeat.Callback<Player> {
```
juankysoriano: Would make sense to move the `Buffer` interface defined here into `Heart.Heartbeat`? So instead of implementing this class, `BufferHeartbeatListener` can implement `Heart.Heartbeat.Buffer` and we can keep this one again as package protected?

core/src/main/java/com/novoda/noplayer/mediaplayer/forwarder/CompletionInfoForwarder.java
Mecharyry: done here c5e691d

core/src/main/java/com/novoda/noplayer/mediaplayer/CheckBufferHeartbeatCallback.java
Mecharyry: You are suggesting moving the `BufferListener` into `Heart.Heartbeat<T>`? I'm not sure it is the responsibility for the `Heart.Heartbeat` to know about this listener, it is only related to buffering hence there being a specific implementation for `bufferHeartbeat`. 

Perhaps I have not understood the benefit this will give us enough.

core/src/main/java/com/novoda/noplayer/mediaplayer/forwarder/CompletionInfoForwarder.java
```diff
@@ -0,0 +1,24 @@
...
+
+        infoListeners.onNewInfo("onCompletion", keyValuePairs);
```
juankysoriano: What's the implementation of `onNewInfo`? Is it also holding internally a `Map<String, String>`? 

I wonder if it makes sense to have different `onNewInfo` methods so we can take rid of the keyvalues? Something like `onNewCompletionInfo(mp)` or any other you need in other info forwarders

core/src/main/java/com/novoda/noplayer/mediaplayer/CheckBufferHeartbeatCallback.java
juankysoriano: The only reason I was seen for this class being public is the fact that we want to access the buffer listener, but given the class is implementing an interface I thought perhaps the buffer listener can be part of that interface it is implemented,

Then we can make this be package protected as it was before

core/src/main/java/com/novoda/noplayer/mediaplayer/forwarder/CompletionInfoForwarder.java
Mecharyry: The problem with doing this is that you would need to have a new `onNewInfo` for each interface in exoplayer and media player that you implement. Which might be overkill when the client will just output these to a Log.

PR #26, opened 2017-06-01
Exoplayer2 avoid leaking dependencies

Dorvaryn:
## Problem

The current implementation uses a few classes from the exoplayer package as public interface.

## Solution

Move the aforementioned classes to the top level package or to the Player interface.

### Test(s) added 

No behaviour changed that requires tests as of now.

### Paired with 

Nobody


Dorvaryn:
There is only 3 more things as far as I know that are leaked from internals.
DRM (will be handled as it's own story), Error reporting(We could create generic errors), PlayerChecker(not sure why or what it is.)

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerTrackSelector.java
```diff
@@ -19,7 +19,7 @@
         for (int i = 0; i < trackInfos.length; i++) {
             MediaPlayer.TrackInfo trackInfo = trackInfos[i];
             if (trackInfo.getTrackType() == MediaPlayer.TrackInfo.MEDIA_TRACK_TYPE_AUDIO) {
-                audioTracks.add(new PlayerAudioTrack(0, i, String.valueOf(trackInfo.hashCode()), trackInfo.getLanguage(), "", -1, -1));
+                audioTracks.add(new PlayerAudioTrack(i, 0, String.valueOf(trackInfo.hashCode()), trackInfo.getLanguage(), "", -1, -1));
```
ouchadam: maybe worth a constant for `0`?

ouchadam:
`PlayerChecker` is used to infer the media player implementation (awesome player/nu player) and then apply a _patch_ to the media player state handling if awesome player is used

PR #28, opened 2017-06-01
Guidelines and Issue Template

Mecharyry:
## Problem
We have an awesome PR template thanks to @rock3r but we don't have contributor guidelines or an issue template 😬 

## Solution
Add the contributor guideline and issue template files.

### Test(s) added 
No, just adding some github files.

### Screenshots
No UI changes.

### Paired with 
Nobody.


.github/contributing.md
```diff
@@ -0,0 +1,42 @@
+# Contributing to Novoda's Open Source projects
```
Mecharyry: Standard from Novoda/base, it should probably be updated soon!

PR #29, opened 2017-06-01
Demo audio track selection

ouchadam:
## Problem

The player supports audio track selecting but it's not present in the demo.

## Solution

To add the ability to switch audio tracks to the demo.

### Test(s) added 

nil tests around the demo code

### Screenshots

| Before | After |
| ------ | ----- |
|device-2017-06-01-162946|device-2017-06-01-162642|

### Paired with 
@Dorvaryn


jackSzm:
Looks good to me ! 💯

PR #30, opened 2017-06-02
ListenersHolder collaborator

Mecharyry:
## Problem
`PlayerListenersHolder` was the super of `ExoPlayerImpl` that held all of the listeners and exposed them through protected method to the `ExoPlayerImpl`. The protected methods made it difficult to test the `ExoPlayerImpl` because we were unable to mock this dependency.

## Solution
Extract `PlayerListenersHolder` as a collaborator so that we can mock it and test more thoroughly the `ExoPlayerImpl`. 

### Test(s) added 
😱 YES! More tests for the `ExoPlayerImpl`.

### Screenshots
No UI changes.

### Paired with 
@juankysoriano and 🍵


core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -242,6 +246,11 @@ public void selectAudioTrack(PlayerAudioTrack audioTrack) {
         return trackSelector.getAudioTracks();
     }
 
+    @Override
+    public PlayerListenersHolder getListenerHolder() {
```
joetimmins: `getListenersHolder`?

Mecharyry:
Yea we considered it too @joetimmins, interfaces are already too large 😢

PR #31, opened 2017-06-02
Replace SimpleExoPlayerView with surface view

ouchadam:
## Problem

We were leaking exoplayer implementation details to all the players as the `SimpleExoPlayerView` was on the `PlayerView` interface. 

## Solution

To replace the `SimpleExoPlayerView` and directly set the `Exoplayer` surface. We now explicitly declare the SurfaceView, Shutter and Subtitles. 

- Also removes `VideoContainer` as it has been replaced with a `ShutterView` which reacts to player state changes instead of us manually calling show/hide

- This means that we lose the built in controls that the `SimpleExoPlayerView` was giving us, spoke with @Dorvaryn and ideally the controls would live separately from the `NoPlayerView, we would need to define this contract.  

### Test(s) added 

Updated the existing ones 😄 

### Screenshots

media controls are removed from the demo

### Paired with 

@rock3r


core/src/main/java/com/novoda/noplayer/VideoContainer.java
```diff
@@ -1,46 +0,0 @@
-package com.novoda.noplayer;
-
-import android.view.View;
-
-import com.novoda.notils.logger.simple.Log;
-
-public class VideoContainer {
```
Mecharyry: 🎉 🎈

core/src/main/res/layout/noplayer_view.xml
```diff
@@ -1,5 +1,25 @@
...
-  android:layout_height="wrap_content" />
+  android:layout_height="wrap_content">
+
+  <SurfaceView
+    android:id="@+id/surface_view"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:layout_gravity="center" />
+
+  <com.google.android.exoplayer2.ui.SubtitleView
```
Mecharyry: Do we want to rely on something from `exo_player`?

rock3r: I think this should be ok for now, we might want to inflate a different layout when using MediaPlayer, but since we have ExoPlayer anyway in the classpath, and given this is unused if we are using MediaPlayer, we should be fine

ouchadam: I'm up for copying the file into the no-player sources as a middleground?

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImplTest.java
```diff
@@ -527,4 +545,13 @@ private MediaSource givenMediaSource() {
     private void givenPlayerIsAttached() {
         player.attach(playerView);
     }
+
+    private SurfaceHolder whenSurfaceHolderIsReady() {
+        ArgumentCaptor<SurfaceHolderRequester.Callback> argumentCaptor = ArgumentCaptor.forClass(SurfaceHolderRequester.Callback.class);
+        verify(surfaceHolderRequester).requestSurfaceHolder(argumentCaptor.capture());
+        SurfaceHolderRequester.Callback callback = argumentCaptor.getValue();
+        SurfaceHolder surfaceHolder = mock(SurfaceHolder.class);
+        callback.onSurfaceHolderReady(surfaceHolder);
```
rock3r: Isn't this logic just repeating what we do in [the `setUp()`](https://github.com/novoda/no-player/blob/7c7215ec420a14819bda9c2846a27e37f6464e97/core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImplTest.java#L140-L146)?

Mecharyry: 😆 yes, I need to learn to read.

core/src/main/res/layout/noplayer_view.xml
Mecharyry: I'm happy with it as is for now. When we get some time out of the current changes we are doing we can implement a no-player version 😄

zegnus: @stefanhoth @ouchadam @Mecharyry can we track this ^ somewhere? :)

Mecharyry: @zegnus I'll add it to my list, I plan to add a project view and start creating issues for things that we need in `no-player`

rock3r: May I suggest having a board [here on GH](https://github.com/novoda/no-player/projects)?

PR #33, opened 2017-06-05
Exoplayer2/ AndroidMediaPlayerFacade tests

Mecharyry:
## Problem
The underlying `AndroidMediaPlayerFacade` is untested 😱 

## Solution
Add some tests so that we can begin tidying up the `AndroidMediaPlayer` flow.

### Test(s) added 
errrr, yes 👅 

### Screenshots
No UI changes.

### Paired with 
Nobody but tea.


core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerAudioTrackSelector.java
```diff
@@ -5,17 +5,18 @@
...
 
 import java.util.ArrayList;
+import java.util.List;
 
 class AndroidMediaPlayerAudioTrackSelector {
 
     private static final int NO_FORMAT = 0;
 
-    ArrayList<PlayerAudioTrack> getAudioTracks(MediaPlayer mediaPlayer) {
+    List<PlayerAudioTrack> getAudioTracks(MediaPlayer mediaPlayer) {
```
Mecharyry: Code to the interface

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -8,27 +8,26 @@
...
+
 class AndroidMediaPlayerFacade {
 
-    private static final int STATE_ERROR = -1;
-    private static final int STATE_IDLE = 0;
-    private static final int STATE_PREPARING = 1;
-    private static final int STATE_PREPARED = 2;
-    private static final int STATE_PLAYING = 3;
-    private static final int STATE_PAUSED = 4;
-    private static final int STATE_PLAYBACK_COMPLETED = 5;
```
Mecharyry: Remove all states! 😈  We now have a `PlaybackState` enum to encapsulate these.

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -97,9 +97,13 @@ private MediaPlayer createMediaPlayer(SurfaceHolder surfaceHolder, Uri videoUri)
         return mediaPlayer;
     }
 
+    protected MediaPlayer createMediaPlayer() {
```
Mecharyry: 🤢  but this is a temporary step so that we can test the behaviour as is, later we will refactor this class.

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -183,7 +187,7 @@ private boolean hasPlayer() {
     }
 
     void start() {
-        if (isInPlaybackState()) {
+        if (playbackStateChecker.isInPlaybackState(mediaPlayer, currentState)) {
```
Mecharyry: Delegate checking of the playback state to a collaborator so that we can simply return `true` or `false` in our tests without needing to setup the media player to be in a certain state.

core/src/main/java/com/novoda/noplayer/mediaplayer/PlaybackStateChecker.java
```diff
@@ -0,0 +1,33 @@
...
+
+import android.media.MediaPlayer;
+
+import static com.novoda.noplayer.mediaplayer.PlaybackStateChecker.PlaybackState.ERROR;
+import static com.novoda.noplayer.mediaplayer.PlaybackStateChecker.PlaybackState.IDLE;
+import static com.novoda.noplayer.mediaplayer.PlaybackStateChecker.PlaybackState.PREPARING;
+
+class PlaybackStateChecker {
+
+    boolean isPlaying(MediaPlayer mediaPlayer, PlaybackState playbackState) {
```
Mecharyry: None of these complex statements are in the `AndroidMediaPlayerFacade` anymore allowing us to simply mock and return either `true` or `false`.

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
```diff
@@ -0,0 +1,482 @@
...
+        givenMediaPlayerIsPrepared();
+
+        ArgumentCaptor<MediaPlayer.OnVideoSizeChangedListener> argumentCaptor = ArgumentCaptor.forClass(MediaPlayer.OnVideoSizeChangedListener.class);
+        verify(mediaPlayer).setOnVideoSizeChangedListener(argumentCaptor.capture());
+        argumentCaptor.getValue().onVideoSizeChanged(mediaPlayer, ANY_WIDTH, ANY_HEIGHT);
+
+        verify(videoSizeChangedListener).onVideoSizeChanged(eq(mediaPlayer), eq(ANY_WIDTH), eq(ANY_HEIGHT));
+    }
+
+    @Ignore("Should we throw when a listener is not attached?")
```
Mecharyry: This is not very well defined in `no-player`, `ExoPlayer` does something different to `AndroidMediaPlayer`

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
```diff
@@ -0,0 +1,482 @@
...
+    public void givenMediaPlayerIsPrepared_whenReleasing_thenReleasesMediaPlayer() {
+        givenMediaPlayerIsPrepared();
+
+        facade.release();
+
+        verify(mediaPlayer).reset();
+        verify(mediaPlayer).release();
+    }
+
+    @Ignore("We should forward to listeners / make logs testable / throw exception.")
```
Mecharyry: Again, this varies depending on the implementation. We should have a clear idea of what to do across both implementations.

core/src/main/java/com/novoda/noplayer/mediaplayer/PlaybackStateChecker.java
```diff
@@ -0,0 +1,33 @@
+package com.novoda.noplayer.mediaplayer;
+
+import android.media.MediaPlayer;
+
+import static com.novoda.noplayer.mediaplayer.PlaybackStateChecker.PlaybackState.ERROR;
+import static com.novoda.noplayer.mediaplayer.PlaybackStateChecker.PlaybackState.IDLE;
+import static com.novoda.noplayer.mediaplayer.PlaybackStateChecker.PlaybackState.PREPARING;
+
+class PlaybackStateChecker {
```
ouchadam: no tests for this class? 😉

Mecharyry: Oh yea 😆 I should probably test this! 💯

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
Dorvaryn: Can we have a comment there then ? Let's make temp code visible and clear.

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
```diff
@@ -0,0 +1,482 @@
...
+                trackSelector,
+                playbackStateChecker
+        ) {
+            @Override
+            protected MediaPlayer createMediaPlayer() {
+                return mediaPlayer;
+            }
+        };
+
+        given(context.getSystemService(Context.AUDIO_SERVICE)).willReturn(audioManager);
```
ouchadam: looks like the `audioService` could/should be a dependency

Mecharyry: done in 5202f23

core/src/main/java/com/novoda/noplayer/mediaplayer/PlaybackStateChecker.java
Mecharyry: TESTS!	e27cfc8

Mecharyry:
@Dorvaryn Can't respond to your comment 😢  but I am tracking with a TODO and with an issue now fd4dbce

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
```diff
@@ -0,0 +1,482 @@
...
+        setShowLogs(false);
+
+        facade = new AndroidMediaPlayerFacade(
+                context,
+                audioManager,
+                trackSelector,
+                playbackStateChecker
+        ) {
+            @Override
+            protected MediaPlayer createMediaPlayer() {
```
juankysoriano: An option could be to have a `MediaPlayerCreator` or something like that, right? that would be the dependency of the class and lazily.

Can be done on a later PR

Mecharyry: yea, exactly 😄  it really depends on how and where we create the `facade` in the `AndroidMediaPlayerImpl`.

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
```diff
@@ -0,0 +1,482 @@
...
+    @Test
+    public void givenNoBoundErrorListener_andMediaPlayerIsPrepared_whenErroring_thenDoesNotForwardError() {
+
+    }
+
+    @Test
+    public void givenBoundBufferListener_andMediaPlayerIsPrepared_whenBuffering_thenBufferPercentageIsUpdated() {
+        givenMediaPlayerIsPrepared();
+
+        ArgumentCaptor<MediaPlayer.OnBufferingUpdateListener> argumentCaptor = ArgumentCaptor.forClass(MediaPlayer.OnBufferingUpdateListener.class);
```
juankysoriano: What about wrapping this into a method `givenMediaPlayerIsBuffering` (for symmetry with the prepared one) ?

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
```diff
@@ -0,0 +1,482 @@
...
+    public void givenMediaPlayerIsNotPrepared_whenStarting_thenNeverSetsDisplay() {
+        facade.start();
+
+        verify(mediaPlayer, never()).setDisplay(surfaceHolder);
+    }
+
+    @Test
+    public void givenMediaPlayerIsPrepared_whenStarting_thenStartsMediaPlayer() {
+        givenMediaPlayerIsPrepared();
+        reset(mediaPlayer);
```
juankysoriano: Why do we have to do the `reset`?

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
```diff
@@ -0,0 +1,482 @@
...
+    @Ignore("We should forward to listeners / make logs testable / throw exception.")
+    @Test
+    public void givenException_whenPreparing_thenLogsCreationError() {
+
+    }
+
+    @Test
+    public void givenBoundPreparedListener_andMediaPlayerIsPrepared_whenPrepared_thenForwardsOnPrepared() {
+        facade.prepareVideo(ANY_URI);
+        ArgumentCaptor<MediaPlayer.OnPreparedListener> argumentCaptor = ArgumentCaptor.forClass(MediaPlayer.OnPreparedListener.class);
```
juankysoriano: Don't we have a `givenMediaPlayerIsPrepared` method doing this?

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
Mecharyry: I was tempted to do that but for me, this is relevant to this test. The one before is done because we need it to occur before this test can be executed. I, personally, dislike switching between methods in a test, I normally lose track of what the test is doing 😄 

Basically, I'm not sure. I like the explicitness of what is being tested as opposed to setup.

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
Mecharyry: This one is actually redundant 😄  thanks for noticing.

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
Mecharyry: We do. This is the explicit test though and I want to avoid hiding what is occurring inside of the method.

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
Mecharyry: There are two resets in this class the second of which is required.

`reset` is used to reset the `mediaPlayer` mock. We reset because in the `givenMediaPlayerIsPrepared` method we will have interactions with `MediaPlayer.setDisplay`, but we aren't interested in those, we are only interested in what occurs outside of it.

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
ouchadam: yep we ended up doing the same thing for the exoplayer impl 👍  (not a blocker for this PR)

PR #38, opened 2017-06-07
Exoplayer2 - Missing Facade Tests

juankysoriano:
## Problem

There was some ignored tests about how to react when the `SurfaceHolderRequester` wasn't set, or any of the many listeners that can be set to the facade.

## Solution

Given that in order to have a correct functioning of media player implementation those listeners / requester has to be set we decided to throw an `IllegalArgumentException` if that's the case. 

Then we completed the missing tests.

Additionally we have created a `MediaPlayerCreator` which is a dependency of `AndroidMediaPlayerFacade`, used to lazily create a `MediaPlayer`. This facilitates testing

### Test(s) added 

Yes, all about tests

### Screenshots

No UI changes

### Paired with 

@Mecharyry

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
@@ -61,8 +61,7 @@ class AndroidMediaPlayerFacade {
 
     void prepareVideo(final Uri videoUri) {
         if (surfaceHolderRequester == null) {
-            logPlayerNotAttachedWarning("prepareVideo()");
-            return;
+            throw new IllegalStateException("Must set a SurfaceHolderRequester before preparing video");
         }
         surfaceHolderRequester.requestSurfaceHolder(new SurfaceHolderRequester.Callback() {
             @Override
@@ -86,7 +85,7 @@ class AndroidMediaPlayerFacade {
 
     private MediaPlayer createAndBindMediaPlayer(SurfaceHolder surfaceHolder, Uri videoUri) throws IOException {
         MediaPlayer mediaPlayer = createMediaPlayer();
-        mediaPlayer.setOnPreparedListener(internalPeparedListener);
+        mediaPlayer.setOnPreparedListener(internalPreparedListener);
         mediaPlayer.setOnVideoSizeChangedListener(internalSizeChangedListener);
         mediaPlayer.setOnCompletionListener(internalCompletionListener);
         mediaPlayer.setOnErrorListener(internalErrorListener);
@@ -117,20 +116,21 @@ class AndroidMediaPlayerFacade {
         @Override
         public void onVideoSizeChanged(MediaPlayer mp, int width, int height) {
             if (onSizeChangedListener == null) {
-                return;
+                throw new IllegalStateException("Should bind a OnVideoSizeChangedListener. Cannot forward events.");
             }
             onSizeChangedListener.onVideoSizeChanged(mp, width, height);
         }
     };
 
-    private final MediaPlayer.OnPreparedListener internalPeparedListener = new MediaPlayer.OnPreparedListener() {
+    private final MediaPlayer.OnPreparedListener internalPreparedListener = new MediaPlayer.OnPreparedListener() {
         @Override
         public void onPrepared(MediaPlayer mp) {
             currentState = PlaybackState.PREPARED;
 
-            if (onPreparedListener != null) {
-                onPreparedListener.onPrepared(mediaPlayer);
+            if (onPreparedListener == null) {
+                throw new IllegalStateException("Should bind a OnPreparedListener. Cannot forward events.");
             }
+            onPreparedListener.onPrepared(mediaPlayer);
         }
     };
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -117,30 +114,32 @@ private void reportCreationError(Exception ex, Uri videoUri) {
         @Override
         public void onVideoSizeChanged(MediaPlayer mp, int width, int height) {
             if (onSizeChangedListener == null) {
-                return;
+                throw new IllegalStateException("Should bind a OnVideoSizeChangedListener. Cannot forward events.");
```
ouchadam: now that we're throwing, it might be good to rename the internal listeners to `forwarders` `OnVideoSizeChangedForwarder` etc

core/src/main/java/com/novoda/noplayer/mediaplayer/MediaPlayerCreator.java
```diff
@@ -0,0 +1,9 @@
+package com.novoda.noplayer.mediaplayer;
+
+import android.media.MediaPlayer;
+
+class MediaPlayerCreator {
```
ouchadam: 👍

ouchadam:
if all the listeners/forwards are not forced, should we create a new public method for attaching them all at once?

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -85,8 +88,8 @@ private void requestAudioFocus() {
     }
 
     private MediaPlayer createAndBindMediaPlayer(SurfaceHolder surfaceHolder, Uri videoUri) throws IOException {
-        MediaPlayer mediaPlayer = createMediaPlayer();
-        mediaPlayer.setOnPreparedListener(internalPeparedListener);
+        MediaPlayer mediaPlayer = mediaPlayerCreator.createMediaPlayer();
+        mediaPlayer.setOnPreparedListener(internalPreparedListener);
```
juankysoriano: @ouchadam when you mean renaming to `forwarder` you mean making this be `internalPreparedForwarder`?

ouchadam: ah no I meant the ones that the internal listeners are delegating 😸

juankysoriano:
@ouchadam for the listeners we are forcing, are 4 of them. I will make a `setListeners` method in the facade that will set all of them

juankysoriano:
@ouchadam comments addressed

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -163,20 +163,11 @@ public void onBufferingUpdate(MediaPlayer mp, int percent) {
...
-        onCompletionListener = listener;
-    }
-
-    void setOnErrorListener(MediaPlayer.OnErrorListener listener) {
-        onErrorListener = listener;
-    }
-
-    void setOnSizeChangedListener(MediaPlayer.OnVideoSizeChangedListener listener) {
-        onSizeChangedListener = listener;
+    void setForwarder(MediaPlayerForwarder forwarder) {
```
ouchadam: 👍

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -61,8 +66,7 @@ void setSurfaceHolderRequester(SurfaceHolderRequester surfaceHolderRequester) {
 
     void prepareVideo(final Uri videoUri) {
         if (surfaceHolderRequester == null) {
-            logPlayerNotAttachedWarning("prepareVideo()");
-            return;
+            throw new IllegalStateException("Must set a SurfaceHolderRequester before preparing video");
```
joetimmins: If it's necessary to set all these things before a AndroidMediaPlayerFacade instance can be used, can we pass them as newInstance parameters?

juankysoriano: Let me check

ouchadam: the surface is attached by a view, it's not guaranteed. Eventually when we support audio only we'll have a separate flow

joetimmins: Does that mean "no"? 😅

juankysoriano: @joetimmins, @ouchadam applies for the surface holder. 

My let me check was more oriented to the listeners. I was taking a look and it would require a bit of redefinition on how we do things when setting up the facade. 

I am open for giving it a try, what about next PR? I remember @Mecharyry mentioning to me that he wanted to keep this PR as "simply unit testing" as possible

joetimmins: Sure, sounds good

PR #39, opened 2017-06-07
ExoPlayer2/Consistent reset logic

ouchadam:
## Problem

The `Player` exposes two ways of releasing the internal resources, `Reset` and `Release` however the exoplayer had never implemented `Reset`.

We also noticed the loading behaviour of the players was inconsistent as the mediaplayer would fully release and recreate before loading new content.

## Solution

Redefine the Release flow of `Player` and makes both `ExoPlayer` and `MediaPlayer` symmetrical.

- Currently `Player` only supports playback when attached to a `PlayerView`

The entire lifecycle is as follows

```
player = createPlayer();
player.attach(playerView);
player.loadVideo('content.mp4');
player.stop(); // Optional as release will have the same effect
player.detach(playerView);
player.release();
``` 

- `Player.reset` has been made private and is called when loading new content to remove the responsibility from clients.

- `Player.stop` is a wrapper for the internal state `reset` to avoid the internal players being in a inconsistent state. 

- Add `clear` to all listener containers

- Adds `clear` to the concrete `ListenersHolder` to clear all bound listeners (part of the new release flow)

- Creates a public `Listeners` interface to only expose add/remove of listeners and not the full `clear`and uses it in the `Player`

- Reintroduces a `ExoPlayerFacade` to keep our exoplayer player related imports in a single place.

- Pushes the `SurfaceHolderRequester` out of the facades and into the `PlayerImpls` as the facades should _only_ be wrapping the 3rd party players and not our domain logic.

- ExoPlayerImpl now throws IllegalStateExceptions when player methods are called before the player is ready, it's now the responsibility of clients to wrap and catch if they do not want this behaviour. (Next PR will add this behaviour to `MediaPlayerImpl`) 

- Refactors the player implementation tests to provide enclosed classes around the player being attached/loaded

- Removes the `Player.onPreRelease` callback, this was a hangover from when the library was within another project. Clients know when release is triggered so they should handle their logic at that point.

- `Player.StateListener.onVideoReleased` has been renamed to `onVideoStopped` and we're triggering the callback when a video stops, stops an existing player and loads, releases and completes.

### Test(s) added 

Yep 

- Impls: release, reset when loading, attach/detach flows.
- `ExoPlayerFacade`
- The moved surface requester logic

### Screenshots

no ui changes

### Paired with 

@rock3r & @Dorvaryn


core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -38,55 +36,48 @@
...
-    void prepareVideo(final Uri videoUri) {
-        if (surfaceHolderRequester == null) {
-            logPlayerNotAttachedWarning("prepareVideo()");
-            return;
+        this.mediaPlayerCreator = mediaPlayerCreator;
+    }
+
+    void prepareVideo(Uri videoUri, SurfaceHolder surfaceHolder) {
+        requestAudioFocus();
+        release();
```
ouchadam: potentially this `release` could be remove as we're doing a `reset` before loading the video if the player has previously loaded

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -116,31 +102,33 @@ private void reportCreationError(Exception ex, Uri videoUri) {
     private final MediaPlayer.OnVideoSizeChangedListener internalSizeChangedListener = new MediaPlayer.OnVideoSizeChangedListener() {
         @Override
         public void onVideoSizeChanged(MediaPlayer mp, int width, int height) {
-            if (onSizeChangedListener == null) {
-                return;
+            if (onVideoSizeChangedForwarder == null) {
+                throw new IllegalStateException("Should bind a OnVideoSizeChangedListener. Cannot forward events.");
```
Dorvaryn: For those 3 it is an invalid state to not have a listener ? This class can't be used if we don't care about those events? Just asking, seems ok but wanted to confirm.

ouchadam: this is from https://github.com/novoda/no-player/pull/38 for now if we only support videos it makes sense to me

juankysoriano: @Dorvaryn good question. 

I had the same debate with @Mecharyry. So basically, in regards to our clients at the moment it is irrelevant if they set the listeners or not. In case they don't the `forwarder` will simply contain an `empty` `Set` of `Listeners`. 

The check is more "internal" for the library. You can consider it a `DeveloperError` and in case reaching that line on the program trace it would mean at some point, somewhere, (concretely `AndroidMediaPlayerImpl`) we forgot to bind to the `forwarder`.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
```diff
@@ -0,0 +1,121 @@
...
+        exoPlayer.setPlayWhenReady(false);
+    }
+
+    public void seekTo(VideoPosition position) {
+        assertVideoLoaded();
+        exoPlayer.seekTo(position.inMillis());
+    }
+
+    public void release() {
+        if (exoPlayer != null) {
```
juankysoriano: Here we are not throwing intentionally, right?

rock3r: Yep

ouchadam: yep! release will invalidate the instance and we're always removing all the listeners.

By not throwing if the instance is null we can keep the client code simpler by not having to maintain the state of being loaded

juankysoriano: I have been checking and I don't see how can we call `release()` being `exoPlayer==null`, but if we want to maintain the check as we do for `stop()` I wonder why not the same mechanism. `assertVideoLoaded()`?

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerFacadeTest.java
```diff
@@ -0,0 +1,319 @@
...
+        @Test
+        public void whenQueryingPlayheadPosition_thenThrowsIllegalStateException() {
+            thrown.expect(IllegalStateException.class);
+
+            facade.getPlayheadPosition();
+        }
+
+        @Test
+        public void whenQueryingMediaDuration_thenThrowsIllegalStateException() {
+            thrown.expect(IllegalStateException.class);
```
juankysoriano: On the `AndroidMediaPlayerFacade` we are asserting the expected thrown with greater granularity. Should we do the same here?

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerFacadeTest.java
```diff
@@ -0,0 +1,319 @@
...
+
+            verify(trackSelector).selectAudioTrack(audioTrack);
+        }
+
+        @Test
+        public void whenGettingAudioTracks_thenDelegatesToTrackSelector() {
+
+            facade.getAudioTracks();
+
+            verify(trackSelector).getAudioTracks();
```
juankysoriano: Please, let's assert that we return the expected value, now that we can. 

```
        given(trackSelector.getAudioTracks(mediaPlayer)).willReturn(AUDIO_TRACKS);

        List<PlayerAudioTrack> audioTracks = facade.getAudioTracks();

        assertThat(audioTracks).isEqualTo(AUDIO_TRACKS);
```

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerFacadeTest.java
```diff
@@ -0,0 +1,319 @@
...
+import org.mockito.junit.MockitoRule;
+
+import static org.fest.assertions.api.Assertions.assertThat;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.inOrder;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+@RunWith(Enclosed.class)
```
juankysoriano: 👍

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -187,9 +203,17 @@ public void pause() {
...
 
     @Override
-    public void loadVideo(Uri uri, ContentType contentType) {
+    public void loadVideo(final Uri uri, ContentType contentType) {
+        if (mediaPlayer.hasPlayedContent()) {
+            reset();
+        }
         listenersHolder.getBufferStateListeners().onBufferStarted();
-        mediaPlayer.prepareVideo(uri);
+        requestSurface(new SurfaceHolderRequester.Callback() {
```
juankysoriano: I wonder if this is a better place to `requestAudioFocus`. Given the symmetry with `requestSurface`. 

Probably out of the scope of this PR

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerFacadeTest.java
ouchadam: 💯 will do

juankysoriano:
LGTM (didn't notice anything weird) 

There are however a couple of comments in tests

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerFacadeTest.java
ouchadam: will do 👍

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
ouchadam: I agree will pulling this out into the impl, but in another PR if that's ok?

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerFacadeTest.java
ouchadam: a3d4e992288b9b8584f055d0dced771e77924ad0

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerFacadeTest.java
ouchadam: c8117c4f48e08c1baebdab6528197d14b7daf230

core/src/test/java/utils/ExceptionMatcher.java
```diff
@@ -1,22 +1,22 @@
-package com.novoda.noplayer.mediaplayer;
+package utils;
```
juankysoriano: 👍

PR #41, opened 2017-06-08
ExoPlayer2/Mediaplayer IllegalStateExeceptions

ouchadam:
## Problem

The `MediaPlayer Facade/Impl` is catching `IllegalStateExceptions` and returning safe values, this is the opposite to what `ExoPlayer` is doing.

## Solution

Sync up the MediaPlayer with ExoPlayer by no longer catching `IllegalStateExceptions` when interacting with the MediaPlayer when it's not in a playback state.

**This is now a responsibility of clients.**

### Test(s) added 

Yep around the exceptions being thrown.

- Also added bonus tests around the `LoadTimeout` 

### Screenshots

no UI changes

### Paired with
 
@Dorvaryn

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/mediaplayer/PlaybackStateChecker.java
+++ b/core/src/main/java/com/novoda/noplayer/mediaplayer/PlaybackStateChecker.java
@@ -27,7 +27,7 @@ class PlaybackStateChecker {
         PREPARED,
         PLAYING,
         PAUSED,
-        COMPLETED;
+        COMPLETED
 
     }
 }
--- a/core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
+++ b/core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
@@ -301,10 +301,10 @@ public class AndroidMediaPlayerFacadeTest {
     }
 
     @Test
-    public void givenMediaPlayerIsNotPrepared_whenStarting_thenNeverSetsDisplay() {
-        facade.start(surfaceHolder);
+    public void givenMediaPlayerIsNotPrepared_whenStarting_thenThrowsIllegalStateException() {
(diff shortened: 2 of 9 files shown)


core/src/main/java/com/novoda/noplayer/mediaplayer/PlaybackStateChecker.java
```diff
@@ -27,7 +27,7 @@ boolean isInPlaybackState(MediaPlayer mediaPlayer, PlaybackState playbackState)
         PREPARED,
         PLAYING,
         PAUSED,
-        COMPLETED;
+        COMPLETED
```
joetimmins: 😂

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
```diff
@@ -381,13 +385,18 @@ public void givenMediaPlayerIsInPlaybackState_whenGettingPosition_thenReturnsPos
     }
 
     @Test
-    public void givenMediaPlayerIsNotInPlaybackState_whenGettingPosition_thenReturnsZero() {
+    public void givenMediaPlayerIsNotInPlaybackState_whenGettingPosition_thenThrowsIllegalStateException() {
+        thrown.expect(
+                ExceptionMatcher.matches(
+                        "Video must be loaded and not in an error state before trying to interact with the player",
```
joetimmins: Can we pull this string out into a field? If it ever changes we'd have to change like three times in the test as it stands

ouchadam: can do 👍

ouchadam: 76d7d3e08d857e20e0b4c578515eaa634945bdbd

PR #43, opened 2017-06-08
Exoplayer2 - Move AndroidMediaPlayerImpl creation + Listeners binding to collaborator

juankysoriano:
## Problem

In a previous PR #38 we discussed about removing the `setForwarder` method from `AndroidMediaPlayerFacade`

## Solution

Done that, what required moving the binding of the listeners done in `AndroidMediaPlayerImpl` to a collaborator class; `AndroidMediaPlayerImplCreator`.

### Test(s) added 

Yes, the `AndroidMediaPlayerImplCreator` is fully tested, moving the tests existing on `AndroidMediaPlayerImpl` to it. 

### Screenshots

No UI changes

### Paired with 

Nobody, but talked with @joetimmins about a potential solution.

**Note** In a future PR (if this succeed) I will do the same for `ExoPlayer2Impl`


core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -175,11 +108,25 @@ private void requestSurface(SurfaceHolderRequester.Callback callback) {
         surfaceHolderRequester.requestSurfaceHolder(callback);
     }
 
+    void resetSeek() {
```
joetimmins: Should these be methods on the `Player` interface?

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -125,6 +126,7 @@ public void onPrepared(MediaPlayer mp) {
         @Override
         public void onCompletion(MediaPlayer mp) {
             currentState = PlaybackState.COMPLETED;
+            MediaPlayer.OnCompletionListener onCompletionForwarder = forwarder.onCompletionListener();
```
joetimmins: Can we enforce the forwarder's listeners to not be null?

joetimmins:
I'm not as up to speed with the `exoplayer` stuff as other people on the project, but this is an improvement IMO on what was there before. Better separation of concerns between configuring the player impl and using it.

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
juankysoriano: It won't be used by any of the clients of `Player`, and very likely not be needed for the other implementation.

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
juankysoriano: If the question is:

`Can the player of the library force this listeners to be null`? The question is no, because in the worst case scenario the forwarder will have a no-op implementation contained. 

But it is possible that we, developers, by mistake do something that causes it hold a `null`. You can consider the thrown below as a `DeveloperError`

joetimmins: Okay cool, just wondering! 👍

ouchadam:
I'm a fan of the separate creator but not totally sold on the setup happening inside of it.

Would prefer that logic to remain inside the impl under a `impl.initialise` method which would be used internally via the `PlayerFactory`

```java
    private Player createMediaPlayer() {
        AndroidMediaPlayerImpl player = androidMediaPlayerImplCreator.create(context);
        player.initialise();
        return player;
    }
```

what do you think?

juankysoriano:
@ouchadam let me check if that remove the original concern: having a `setForwarder` method on the facade

juankysoriano:
@ouchadam 819503229db6cc0c2ba69bb5b51dc35b2b3ff900

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -24,36 +24,35 @@
     private static final Map<String, String> NO_HEADERS = null;
 
     private final Context context;
+    private MediaPlayerForwarder forwarder;
```
ouchadam: this can be final 😸

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -47,46 +45,35 @@
...
         this.listenersHolder = listenersHolder;
+        this.bufferHeartbeatCallback = bufferHeartbeatCallback;
         this.loadTimeout = loadTimeoutParam;
         this.heart = heart;
         this.handler = handler;
         this.buggyVideoDriverPreventer = buggyVideoDriverPreventer;
-        heart.bind(new Heart.Heartbeat<>(listenersHolder.getHeartbeatCallbacks(), this));
+    }
 
+    public void initialise() {
```
ouchadam: 👍

ouchadam:
I like it! It's probably worth making a similar change to ExoPlayerImpl as well in another PR 👻

juankysoriano:
Yes I will do it in exoplayer!

PR #44, opened 2017-06-09
Exoplayer2 - Decouple binding from constructor in ExoPlayerImpl

juankysoriano:
## Problem

As we did on #43 we want to decouple the binding from the constructor on `ExoPlayerTwoImpl`

## Solution

Create an `initialise()` method that will be doing the binding. The instantiation of the class is now done in a factory instead of a `newInstance()`. `initialise()` is called upon creation, on `PlayerFactory`. 

### Test(s) added 

Yes, modified existing ones and made sure that initialisation is happening when it should

### Screenshots

No UI changes

### Paired with 

Nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
+++ b/core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
@@ -23,7 +23,7 @@ import com.novoda.noplayer.player.PlayerInformation;
 
 import java.util.List;
 
-public final class AndroidMediaPlayerImpl implements Player {
+public class AndroidMediaPlayerImpl implements Player {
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -23,7 +23,7 @@
 
 import java.util.List;
 
-public final class AndroidMediaPlayerImpl implements Player {
+public class AndroidMediaPlayerImpl implements Player {
```
joetimmins: Why no more `final`? I like `final`! 😅

joetimmins: Oh because we need to mock it....

PR #45, opened 2017-06-09
ExoPlayer 2 - Fix build by removing notils

juankysoriano:
## Problem

We are having a lot of issues because of the `notils` support library versions clashing with libraries support versions.

## Solution

We are not using notils a lot, instead I have "ported" the Logging function to an `utils.Log` class

### Test(s) added 

Modified some existing tests 

### Screenshots

No UI changes

### Paired with 

Nobody

### DIFF
--- a/core/src/main/java/com/novoda/utils/Log.java
+++ b/core/src/main/java/com/novoda/utils/Log.java
@@ -1,0 +1,136 @@
+package com.novoda.utils;
+
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+public abstract class Log {
+
+    private static final int DEPTH = 5;
+    private static final int CLASS_SUFFIX = 5;
+
+    private static boolean isEnabled = true;
+
+    public static void setLoggingEnabled(boolean enabled) {
+        isEnabled = enabled;
+    }
+
+    private static String logMessage(String message, Throwable throwable) {
+        String detailedMessage = getDetailedLog(message);
+        if (throwable != null) {
+            detailedMessage += "\n" + getStackTraceString(throwable);
+        }
+        return detailedMessage;
+    }
+
+    private static String getDetailedLog(String message) {
+        Thread current = Thread.currentThread();
+        final StackTraceElement trace = current.getStackTrace()[DEPTH];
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/novoda/utils/Log.java
```diff
@@ -0,0 +1,136 @@
+package com.novoda.utils;
+
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+public abstract class Log {
```
ouchadam: we should rename this so that clients of the library don't get spammed with `Log` classes 🍡 `NoPlayerLog`?

juankysoriano: 👍

juankysoriano: I was going to remove it, named it `Log` to make it easier the migration and forgot to rename it afterwards. Good point.

core/src/main/java/com/novoda/utils/NoPlayerLog.java
```diff
@@ -0,0 +1,136 @@
...
+            return;
+        }
+        android.util.Log.e("No-Player", logMessage(msg, null));
+    }
+
+    public static void e(Throwable throwable, String msg) {
+        if (!isEnabled) {
+            return;
+        }
+        android.util.Log.e("No-Player", logMessage(msg, throwable));
```
rock3r: Nit extract constant for the tag?

core/src/main/java/com/novoda/utils/NoPlayerLog.java
```diff
@@ -0,0 +1,136 @@
+package com.novoda.utils;
+
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+public abstract class NoPlayerLog {
```
juankysoriano: This is basically a quick port of what we have in `notils`

core/src/main/java/com/novoda/utils/NoPlayerLog.java
juankysoriano: sure

PR #46, opened 2017-06-09
Exoplayer2/Streaming & Local widevine modular drm support

ouchadam:
## Problem

We removed the widevine modular streaming and download support whilst migrating to ExoPlayer2.

## Solution

- Adds a widevine modular example to the demo.

- Passes a `DrmSessionCreator` to the ExoPlayer `RenderersFactory` to enable DRM content. 

- Wraps the `Key` and `Provision` Drm requests to avoid exposing internal api changes. 

- Reintroduces the `LocalDrmSession` but ported to `ExoPlayer2`'s `DrmSessionManager` and `DrmSession`

- TODO propagate errors via an event listener inside the `LocalDrmSession`
- TODO handle errors in the `StreamingDrmSessionCreator`

### Test(s) added 

updated the existing but tests should be written around these classes, but I've gotta bounce 🌴 🌞 

### Paired with 

nobody, 🌴 🌞

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
@@ -97,6 +97,7 @@ class LocalDrmSession implements DrmSessionManager<FrameworkMediaCrypto>, DrmSes
     }
 
     private void onError(Exception e) {
+        // TODO add listener and callback with exception see DefaultDrmSessionManager onError
         state = STATE_ERROR;
         lastKnownException = e;
     }
--- a/core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
@@ -15,6 +15,8 @@ import com.google.android.exoplayer2.drm.FrameworkMediaDrm;
 import com.google.android.exoplayer2.drm.UnsupportedDrmException;
 
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 
@@ -68,7 +70,7 @@ class LocalDrmSession implements DrmSessionManager<FrameworkMediaCrypto>, DrmSes
 
     @Override
     public Map<String, String> queryKeyStatus() {
-        return null;
+        return Collections.unmodifiableMap(new HashMap<String, String>());
     }
 
     @Override
--- a/core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
@@ -1,0 +1,103 @@
+package com.novoda.noplayer.drm;
+
+import android.annotation.TargetApi;
+import android.media.MediaCryptoException;
+import android.media.MediaDrmException;
+import android.os.Build;
+import android.os.Looper;
+
+import com.google.android.exoplayer2.drm.DrmInitData;
+import com.google.android.exoplayer2.drm.DrmSession;
+import com.google.android.exoplayer2.drm.DrmSessionManager;
+import com.google.android.exoplayer2.drm.ExoMediaDrm;
+import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
+import com.google.android.exoplayer2.drm.FrameworkMediaDrm;
+import com.google.android.exoplayer2.drm.UnsupportedDrmException;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.UUID;
+
+@TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR2)
+public class LocalDrmSession implements DrmSessionManager<FrameworkMediaCrypto>, DrmSession<FrameworkMediaCrypto> {
+
+    private static final UUID WIDEVINE_MODULAR_UUID = new UUID(0xEDEF8BA979D64ACEL, 0xA3C827DCD51D21EDL);
+
+    private final ExoMediaDrm<FrameworkMediaCrypto> mediaDrm;
+    private final byte[] keySetIdToRestore;
+
+    private int state = STATE_CLOSED;
+    private byte[] sessionId;
+    private FrameworkMediaCrypto mediaCrypto;
+    private Exception lastKnownException;
+
+    public LocalDrmSession(byte[] keySetIdToRestore) throws UnsupportedDrmException {
+        this.keySetIdToRestore = Arrays.copyOf(keySetIdToRestore, keySetIdToRestore.length);
+        this.mediaDrm = FrameworkMediaDrm.newInstance(WIDEVINE_MODULAR_UUID);
+    }
+
+    @Override
+    public int getState() {
+        return state;
+    }
+
+    @Override
+    public FrameworkMediaCrypto getMediaCrypto() {
+        if (state != STATE_OPENED && state != STATE_OPENED_WITH_KEYS) {
+            throw new IllegalStateException();
+        }
+        return mediaCrypto;
+    }
+
+    @Override
+    public boolean requiresSecureDecoderComponent(String mimeType) {
+        if (state != STATE_OPENED && state != STATE_OPENED_WITH_KEYS) {
+            throw new IllegalStateException();
+        }
+        return mediaCrypto.requiresSecureDecoderComponent(mimeType);
+    }
+
+    @Override
+    public DrmSessionException getError() {
+        return new DrmSessionException(lastKnownException);
+    }
+
+    @Override
+    public Map<String, String> queryKeyStatus() {
+        return null;
+    }
+
+    @Override
+    public byte[] getOfflineLicenseKeySetId() {
+        return keySetIdToRestore;
+    }
+
+    @Override
+    public DrmSession<FrameworkMediaCrypto> acquireSession(Looper playbackLooper, DrmInitData drmInitData) {
+        try {
+            state = STATE_OPENING;
+            sessionId = mediaDrm.openSession();
+            mediaCrypto = mediaDrm.createMediaCrypto(WIDEVINE_MODULAR_UUID, sessionId);
+            state = STATE_OPENED;
+
+            mediaDrm.restoreKeys(sessionId, keySetIdToRestore);
+            state = STATE_OPENED_WITH_KEYS;
+        } catch (MediaDrmException | MediaCryptoException e) {
+            onError(e);
+        }
+        return this;
+    }
+
+    @Override
+    public void releaseSession(DrmSession<FrameworkMediaCrypto> drmSession) {
+        if (sessionId != null) {
+            mediaDrm.closeSession(sessionId);
+            sessionId = null;
+        }
+    }
+
+    private void onError(Exception e) {
+        state = STATE_ERROR;
+        lastKnownException = e;
+    }
+}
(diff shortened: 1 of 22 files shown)


core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
```diff
@@ -0,0 +1,104 @@
...
+        } catch (MediaDrmException | MediaCryptoException e) {
+            onError(e);
+        }
+        return this;
+    }
+
+    @Override
+    public void releaseSession(DrmSession<FrameworkMediaCrypto> drmSession) {
+        if (sessionId != null) {
+            mediaDrm.closeSession(sessionId);
```
juankysoriano: I wonder if we should make ` state` here be `STATE_CLOSED` ?

core/src/main/java/com/novoda/noplayer/drm/StreamingDrmSessionCreator.java
```diff
@@ -0,0 +1,46 @@
...
+    private final MediaDrmCallback mediaDrmCallback;
+
+    public StreamingDrmSessionCreator(MediaDrmCallback mediaDrmCallback) {
+        this.mediaDrmCallback = mediaDrmCallback;
+    }
+
+    @Override
+    public DrmSessionManager<FrameworkMediaCrypto> create() {
+        Looper eventLooper = Looper.myLooper() != null ? Looper.myLooper() : Looper.getMainLooper();
+        Handler mainHandler = new Handler(eventLooper);
```
juankysoriano: just `handler`

core/src/main/java/com/novoda/noplayer/drm/StreamingDrmSessionCreator.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.android.exoplayer2.drm.UnsupportedDrmException;
+
+import java.util.HashMap;
+import java.util.UUID;
+
+public class StreamingDrmSessionCreator implements DrmSessionCreator {
+
+    private static final UUID WIDEVINE_MODULAR_UUID = new UUID(0xEDEF8BA979D64ACEL, 0xA3C827DCD51D21EDL);
+    private static final HashMap<String, String> NO_OPTIONAL_PARAMETERS = null;
+    private static final DefaultDrmSessionManager.EventListener TODO_ERROR_EVENT_LISTENER = null;
```
juankysoriano: 💯

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerCreator.java
```diff
@@ -18,8 +19,8 @@
         this.trackSelector = trackSelector;
     }
 
-    public SimpleExoPlayer create() {
-        DefaultRenderersFactory renderersFactory = new DefaultRenderersFactory(context);
+    public SimpleExoPlayer create(DrmSessionCreator drmSessionCreator) {
```
juankysoriano: ` DrmSessionCreator` could as well be a dependency of this class, right?

juankysoriano:
I has been agreed that we'll hold on about DRM until @ouchadam is back. For the next Sprint we'll be focusing on subtitles

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public DrmSessionException getError() {
+        return new DrmSessionException(lastKnownException);
+    }
+
+    @Override
+    public Map<String, String> queryKeyStatus() {
+        return null;
```
rock3r: Shouldn't this return an `unmodifiableMap()` or the method be annotated as `@Nullable`?

core/src/main/java/com/novoda/noplayer/drm/NoDrmSessionCreator.java
```diff
@@ -1,5 +1,12 @@
 package com.novoda.noplayer.drm;
 
+import com.google.android.exoplayer2.drm.DrmSessionManager;
+import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
+
 public class NoDrmSessionCreator implements DrmSessionCreator {
 
+    @Override
+    public DrmSessionManager<FrameworkMediaCrypto> create() {
+        return null;
```
rock3r: Ditto, should this be marked `@Nullable`?

core/src/main/java/com/novoda/noplayer/drm/StreamingDrmSessionCreator.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final MediaDrmCallback mediaDrmCallback;
+
+    public StreamingDrmSessionCreator(MediaDrmCallback mediaDrmCallback) {
+        this.mediaDrmCallback = mediaDrmCallback;
+    }
+
+    @Override
+    public DrmSessionManager<FrameworkMediaCrypto> create() {
+        Looper eventLooper = Looper.myLooper() != null ? Looper.myLooper() : Looper.getMainLooper();
```
rock3r: Is this implicit threading model necessary? I'm always a bit weary of implicit `Handler` hopping, especially when it makes testing impossible (like in this case)

core/src/main/java/com/novoda/noplayer/drm/StreamingDrmSessionCreator.java
```diff
@@ -0,0 +1,46 @@
...
+            return new DefaultDrmSessionManager<>(
+                    WIDEVINE_MODULAR_UUID,
+                    frameworkMediaDrm,
+                    mediaDrmCallback,
+                    NO_OPTIONAL_PARAMETERS,
+                    mainHandler,
+                    TODO_ERROR_EVENT_LISTENER
+            );
+        } catch (UnsupportedDrmException e) {
+            throw new RuntimeException();
```
rock3r: This should really _not_ replace a specific exception with a generic one, without even passing in the cause. I suspect we are not running static analysis on this repo, it'd have flagged this!

core/src/main/java/com/novoda/noplayer/drm/provision/ModularDrmKeyRequest.java
```diff
@@ -0,0 +1,20 @@
...
+
+    private final String url;
+    private final byte[] data;
+
+    public ModularDrmKeyRequest(String url, byte[] data) {
+        this.url = url;
+        this.data = data;
+    }
+
+    public String getUrl() {
```
rock3r: `url()`? same below

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerCreator.java
rock3r: It seems like it's injected at a later stage than this class is initialised, upstream

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
```diff
@@ -0,0 +1,104 @@
...
+    public FrameworkMediaCrypto getMediaCrypto() {
+        if (state != STATE_OPENED && state != STATE_OPENED_WITH_KEYS) {
+            throw new IllegalStateException();
+        }
+        return mediaCrypto;
+    }
+
+    @Override
+    public boolean requiresSecureDecoderComponent(String mimeType) {
+        if (state != STATE_OPENED && state != STATE_OPENED_WITH_KEYS) {
```
Mecharyry: Same as above, could extract this.

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
Mecharyry: I'd vote for `unmodifiableMap` here over the annotation.

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
```diff
@@ -0,0 +1,104 @@
...
+import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
+import com.google.android.exoplayer2.drm.FrameworkMediaDrm;
+import com.google.android.exoplayer2.drm.UnsupportedDrmException;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.UUID;
+
+@TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR2)
+class LocalDrmSession implements DrmSessionManager<FrameworkMediaCrypto>, DrmSession<FrameworkMediaCrypto> {
```
Mecharyry: I'm confused because it seems like this should be two different objects. Should we not have a manager, responsible for acquiring a session, and a `DrmSession`, from which you can retrieve things like `FrameworkMediaCrypto`. Why were these originally coupled together?

core/src/main/java/com/novoda/noplayer/drm/StreamingDrmSessionCreator.java
Mecharyry: I guess this was done because we are passing a constant to the call that would crash but yea, we should not through a generic.

core/src/main/java/com/novoda/noplayer/drm/provision/ModularDrmKeyRequest.java
Mecharyry: Also, add `toString` and `hashcode` I don't like reading gibberish when I get a test crash 😄

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerCreator.java
Mecharyry: It should be fine to move as @juankysoriano says. We are creating it in the `ExoPlayerTwoImplFactory` and it is only passed through the constructor, stored and use only for the internal `loadVideo`. We don't give an option to the client to pass through a different `DrmSessionCreator` we need to create from the `ExoPlayerTwoImplFactory`.

demo/src/main/java/com/novoda/demo/DataPostingModularDrm.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.android.exoplayer2.upstream.DefaultHttpDataSourceFactory;
+import com.google.android.exoplayer2.upstream.HttpDataSource;
+import com.google.android.exoplayer2.util.Util;
+import com.novoda.noplayer.drm.DrmRequestException;
+import com.novoda.noplayer.drm.StreamingModularDrm;
+import com.novoda.noplayer.drm.provision.ModularDrmKeyRequest;
+
+import java.io.IOException;
+
+class DataPostingModularDrm implements StreamingModularDrm {
```
Mecharyry: `StreamingModularDrm` is only used by this class, can we have it implement the `DrmHandler` instead? Same with `DownloadedModularDrm`?

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
rock3r: Me too 💃

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
ouchadam: follows from the default implementation, https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/drm/DefaultDrmSessionManager.java

The session creation is coupled to the drm manager lifecycle

Mecharyry: urgh gross 😄  Oh well!

Mecharyry:
We started again to make sure others had knowledge.

PR #47, opened 2017-06-16
ALL-3995/Add subtitles functionality

joetimmins:
## Problem
We weren't showing subtitles in our NoPlayer demo app.

## Solution
We added the concept of a `PlayerSubtitlesView` as an interface, so we had something the player could call back to in order to show subtitles.

We added methods to our `Player` interface to get the subtitle tracks from a video, show a specific subtitle track, or show just the first available track. We also added a method to hide subtitles.

We have also provided dynamic indexing for all renderer tracks. Before, we were hardcoding the track at position 0 to always be audio.

### Test(s) added 
We tested the `ExoPlayerImpl` class around the new subtitles methods we added.

### Screenshots
| After |
| ----- |
| subtitles-on |

### Paired with 

@zegnus


core/src/main/java/com/novoda/noplayer/PlayerSubtitlesView.java
```diff
@@ -0,0 +1,14 @@
+package com.novoda.noplayer;
+
+import com.google.android.exoplayer2.text.Cue;
+
+import java.util.List;
+
+public interface PlayerSubtitlesView {
```
juankysoriano: What about renaming to `SubtiteableView` and making `PlayerView extend SubtiteableView`?

juankysoriano: Otherwise I think this is still ok, but maybe with a renaming, `PlayerSubtitlesView` sounds weird to me (maybe just me)

core/src/main/java/com/novoda/noplayer/Player.java
```diff
@@ -33,8 +33,18 @@
 
     void selectAudioTrack(PlayerAudioTrack audioTrack);
 
+    boolean hasAvailableSubtitles();
```
juankysoriano: Do we have an equivalent method for `audioTrack`?

core/src/main/java/com/novoda/noplayer/Player.java
```diff
@@ -33,8 +33,18 @@
...
+
+    void selectSubtitleTrack(PlayerSubtitleTrack subtitleTrack, PlayerSubtitlesView playerSubtitlesView);
+
+    void selectFirstAvailableSubtitleTrack(PlayerSubtitlesView playerSubtitlesView);
```
juankysoriano: Do we have an equivalent method for `audioTrack`? If we don't how do we select the first available one for audio? We should push for symmetry

core/src/main/java/com/novoda/noplayer/Player.java
```diff
@@ -33,8 +33,18 @@
...
+
+    void clearSubtitleTrack(PlayerSubtitlesView playerSubtitlesView);
```
juankysoriano: What would be an use case for this?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -37,12 +44,14 @@
...
                      LoadTimeout loadTimeoutParam,
-                     Heart heart) {
+                     Heart heart,
+                     ExoPlayerTrackSelector exoPlayerTrackSelector) {
         this.exoPlayer = exoPlayer;
         this.listenersHolder = listenersHolder;
         this.loadTimeout = loadTimeoutParam;
         this.forwarder = exoPlayerForwarder;
         this.heart = heart;
+        this.exoPlayerTrackSelector = exoPlayerTrackSelector;
```
Dorvaryn: Why do we have to know about the exoPlayerTrackSelector here ? Shouldn't we delegate to exoPlayerFacade who already deals with this ?

core/src/main/java/com/novoda/noplayer/Player.java
zegnus: will rename to `hide...`

joetimmins: This is invoked when the user wants to turn subtitles off

core/src/main/java/com/novoda/noplayer/PlayerSubtitlesView.java
jackSzm: What I don't like about this interface that it is supposed to be used on both implementation of Player but it is coupled with exoplayer. Maybe we could find a way to pass something else than `com.google.android.exoplayer2.text.Cue`

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -196,11 +212,50 @@ public void selectAudioTrack(PlayerAudioTrack audioTrack) {
     }
 
     @Override
+    public boolean hasAvailableSubtitles() {
+        return !exoPlayer.getSubtitleTracks().isEmpty();
+    }
+
+    @Override
+    public void selectSubtitleTrack(PlayerSubtitleTrack subtitleTrack, final PlayerSubtitlesView playerSubtitlesView) {
```
Dorvaryn: Instead of passing the view to every method using it should we use a similar approach to the playerView and use attach/detach ? To be consistent with out API

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -58,6 +67,13 @@ public void initialise() {
             @Override
             public void onPrepared(PlayerState playerState) {
                 loadTimeout.cancel();
+
+                RendererTrackIndexExtractor rendererTrackIndexExtractor = new RendererTrackIndexExtractor();
+                Map<TrackType, Integer> trackTypeIndexMap = rendererTrackIndexExtractor.extractFrom(
+                        exoPlayerTrackSelector.trackInfo(),
+                        exoPlayer.getRawExoPlayer()
+                );
+                exoPlayerTrackSelector.setTrackRendererIndexes(trackTypeIndexMap);
```
Dorvaryn: This looks like it could be delegated as mentioned above ?

core/src/main/java/com/novoda/noplayer/Player.java
zegnus: I think that the current impl assumes that there is always an audio track

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
```diff
@@ -104,18 +114,52 @@ public void loadVideo(Uri uri, ContentType contentType, ExoPlayerForwarder forwa
...
+        return audioTrackSelector.getAudioTracks();
+    }
+
+    public void setSubtitleRendererOutput(TextRenderer.Output output) {
+        this.output = output;
+        setExoPlayerTextOutput(output);
+    }
+
+    private void setExoPlayerTextOutput(TextRenderer.Output output) {
+        if (exoPlayer == null) {
```
juankysoriano: shouldn't we do `assertVideoLoaded` instead? look at other usages of that method

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/ExoPlayerMappedTrackInfo.java
```diff
@@ -0,0 +1,25 @@
+package com.novoda.noplayer.exoplayer.mediasource;
+
+import com.google.android.exoplayer2.source.TrackGroupArray;
+import com.google.android.exoplayer2.trackselection.MappingTrackSelector;
+
+public class ExoPlayerMappedTrackInfo {
```
Dorvaryn: We need to be careful with public classes, do we need to expose this to clients of the library ?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
```diff
@@ -104,18 +114,52 @@ public void loadVideo(Uri uri, ContentType contentType, ExoPlayerForwarder forwa
...
+
+    public List<PlayerSubtitleTrack> getSubtitleTracks() {
+        return subtitleTrackSelector.getSubtitleTracks();
     }
 
     public boolean hasPlayedContent() {
         return exoPlayer != null;
     }
+
+    public SimpleExoPlayer getRawExoPlayer() {
```
juankysoriano: This is breaking the full intention of the facade, why do we need to get the raw exoplayer?

zegnus: Yeah, I'm looking at it right now

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -211,6 +213,11 @@ void stop() {
         return trackSelector.getAudioTracks(mediaPlayer);
     }
 
+    public List<PlayerSubtitleTrack> getSubtitleTracks() {
+        NoPlayerLog.w("Subtitles not implemented for Android Media Player");
```
Dorvaryn: Should we be more aggressive than a log here ?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
juankysoriano: If you need something from exoplayer let the facade provide it. Please use the facade as a facade

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -264,11 +267,37 @@ public void selectAudioTrack(PlayerAudioTrack audioTrack) {
     }
 
     @Override
+    public boolean hasAvailableSubtitles() {
+        NoPlayerLog.w("Subtitles not implemented for Android Media Player");
```
Dorvaryn: Same on all of those, wondering if a log is clear enough.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
zegnus: That method will throw an exception, we don't want that here

core/src/main/java/com/novoda/noplayer/PlayerSubtitleTrack.java
```diff
@@ -0,0 +1,107 @@
+package com.novoda.noplayer;
+
+public class PlayerSubtitleTrack {
```
Dorvaryn: This is a public class we give to our clients, I think a lot of this data might never exist in MediaPlayer if we ever implement it. We should probably restrict ourselves to the minimum for now or check what is available. It's easier to add things to a public API than to remove it.

core/src/main/java/com/novoda/noplayer/Player.java
Dorvaryn: Wouldn't using `getSubtitleTracks ` and checking for empty enough ?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -196,11 +212,50 @@ public void selectAudioTrack(PlayerAudioTrack audioTrack) {
...
+
+    @Override
+    public void selectFirstAvailableSubtitleTrack(PlayerSubtitlesView playerSubtitlesView) {
+        setSubtitleRenderer(playerSubtitlesView);
+        exoPlayer.selectFirstAvailableSubtitlesTrack();
+        playerSubtitlesView.showSubtitles();
+    }
+
+    private void setSubtitleRenderer(final PlayerSubtitlesView playerSubtitlesView) {
+        exoPlayer.setSubtitleRendererOutput(new TextRenderer.Output() {
```
juankysoriano: Is it ok to pass the `PlayerSubtitlesView` as a dependency of this methods? we already have an `attach` method and I wonder if this `setSubtitleRenderer` can be done there... 

I can also appreciate that this class do very little direct interaction with `PlayerView` and does stuff on it via listeners, so I wonder if we should be doing the same.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
juankysoriano: Is it possible we try to set this before the video is loaded? in what scenario?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
juankysoriano: I agree with @Dorvaryn , let's use the facade

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/ExoPlayerMappedTrackInfo.java
juankysoriano: Why does this class have to be `public`? I suspect it is only for testing, can we revisit?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
```diff
@@ -104,18 +114,52 @@ public void loadVideo(Uri uri, ContentType contentType, ExoPlayerForwarder forwa
...
-        trackSelector.selectAudioTrack(audioTrack);
+        audioTrackSelector.selectAudioTrack(audioTrack);
     }
 
     public List<PlayerAudioTrack> getAudioTracks() {
-        return trackSelector.getAudioTracks();
+        return audioTrackSelector.getAudioTracks();
+    }
+
+    public void setSubtitleRendererOutput(TextRenderer.Output output) {
```
juankysoriano: Can we tests this new methods in the facade?

core/src/main/java/com/novoda/noplayer/exoplayer/TrackType.java
```diff
@@ -0,0 +1,7 @@
+package com.novoda.noplayer.exoplayer;
+
+public enum TrackType {
```
juankysoriano: anyway we can make this be `package`?

core/src/main/java/com/novoda/noplayer/Player.java
zegnus: yup, will remove

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
juankysoriano: Maybe a warning is enough but I would maybe make sure there is a `Unit` test written for this even if they are dummy tests for now. For example here asserting we return false, or in the others verifying no interaction with the input arguments

The reason is that we devoted a good time writing test for this and it is proven that we might easily forget about maintaining them. What do you think guys?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
zegnus: You could call `showFirstAvailableSubtitleTrack` before loading a video as is a public method

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
zegnus: I think that there was a instantiation dependency problem and this is why we did it this way, but I'm going to have a look again to the whole bit and make it better 👍

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
juankysoriano: And why would you do that? If you do that you deserve an exception because the video is not loaded.

That a method is `public` doesn't mean that you have the right to use out of the rules of the API, isn't it?

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
Dorvaryn: I was kind off thinking to communicate to our library clients. Maybe it should throw ? As in is there a valid reason to support those methods being called ?

juankysoriano: I would agree with you if this thing were not under a `Player` interface, in which case maybe the client doesn't even have control... 

I am still hesitant to be honest, in one sense I agree with you.

joetimmins: I don't think it should throw. Both these players adhere to the `Player` interface and it feels like we'd be leaking knowledge of which player the client is actually using if we had a thrown exception here.

If the client already knows which media player they're using, maybe we shouldn't have them both adhering to the same interface?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
joetimmins: I think "deserve" an exception is a bit strong here! I like being able to call the methods in any order, it doesn't really add that much complexity and it would make it simpler for users of the library to not have to care about the order in which they call methods like this.

core/src/main/java/com/novoda/noplayer/Player.java
Mecharyry: I agree. We should push for symmetry, I'd rather avoid having lots of helper methods that is going to make the Player interface bloat.

core/src/main/java/com/novoda/noplayer/PlayerSubtitlesView.java
Mecharyry: I agree. We should avoid using `ExoPlayer` or `MediaPlayer` specific code in the `no-player` part of this library.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
```diff
@@ -104,18 +114,52 @@ public void loadVideo(Uri uri, ContentType contentType, ExoPlayerForwarder forwa
                 forwarder.mediaSourceEventListener()
         );
         exoPlayer.prepare(mediaSource, RESET_POSITION, DO_NOT_RESET_STATE);
+        setExoPlayerTextOutput(output);
     }
 
     public void selectAudioTrack(PlayerAudioTrack audioTrack) {
```
Mecharyry: Can you boyscout all the `public` methods here and make them package?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
Mecharyry: I agree with @juankysoriano. We should throw in here. It is better to explicitly throw and tell your clients that they have called this method at the wrong time, than for them to assume that it should work but actually it doesn't because we quick return.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
Mecharyry: Thanks for catching this @juankysoriano 💯  We should never expose the underlying `ExoPlayer` it defeats the point of having the `Player` interface and the abstraction of this library. Happy to help refactor this out @zegnus

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
Mecharyry: Ah someone else has said this 😆

core/src/main/java/com/novoda/noplayer/exoplayer/TrackType.java
Mecharyry: Yes. If we move the `RendererTrackIndexExtractor` to the `mediasource` package, along with this, we should be able to make this package.

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/ExoPlayerTrackSelector.java
```diff
@@ -4,36 +4,60 @@
...
-    boolean supportsTrackSwitching(TrackGroupArray trackGroups, int groupIndex) {
+    boolean supportsTrackSwitching(TrackType trackType, TrackGroupArray trackGroups, int groupIndex) {
+        Integer audioRendererIndex = rendererTrackIndex.get(trackType);
         return trackGroups.get(groupIndex).length > 0
-                && trackInfo().getAdaptiveSupport(TRACK_TYPE_AUDIO, groupIndex, false) != RendererCapabilities.ADAPTIVE_NOT_SUPPORTED;
+                && trackInfo().getAdaptiveSupport(audioRendererIndex, groupIndex, false) != RendererCapabilities.ADAPTIVE_NOT_SUPPORTED;
+    }
+
+    public void setTrackRendererIndexes(Map<TrackType, Integer> trackTypeIndexMap) {
+        rendererTrackIndex = new EnumMap<>(trackTypeIndexMap);
```
Mecharyry: Why are we forced to pass the map through here instead of the `constructor`?

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
Mecharyry: I don't think we should have any logs in the library if I am completely honest. We should either forward to one of the `Player` listeners that will notify the client or throw. We should avoid this form of `Log` which is at odds with how the library operates.

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
Mecharyry: I think this should throw and the issue is actually in the `Player` factory and how it selects a `Player`. Perhaps the client should have a list of requirements of the `Player` they want and the `Factory` will select the most appropriate, similar to how it operates for `DRM`.

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImplTest.java
```diff
@@ -424,6 +433,62 @@ public void whenStartingPlayAtVideoPosition_thenNotifiesStateListenersThatVideoI
...
+        }
+
+        private List<Cue> givenPlayerHasLoadedSubtitleCues() {
+            final List<Cue> cueList = Arrays.asList(new Cue("first cue"), new Cue("secondCue"));
+            doAnswer(new Answer() {
+                @Override
+                public Object answer(InvocationOnMock invocation) throws Throwable {
+                    TextRenderer.Output output = invocation.getArgument(0);
+                    output.onCues(cueList);
+                    return null;
```
Mecharyry: When you are returning `null` you can use `Void` I believe.

core/src/test/java/com/novoda/noplayer/exoplayer/PlayerSubtitleTrackFixture.java
```diff
@@ -0,0 +1,61 @@
+package com.novoda.noplayer.exoplayer;
+
+import com.novoda.noplayer.PlayerSubtitleTrack;
+
+public class PlayerSubtitleTrackFixture {
```
Mecharyry: All `Fixtures` should be package to avoid leaking these to the clients.

core/src/test/java/com/novoda/noplayer/exoplayer/RendererTrackIndexExtractorTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public void setUp() throws Exception {
+        extractor = new RendererTrackIndexExtractor();
+        given(trackCounter.length()).willReturn(1);
+    }
+
+    @Test
+    public void givenAudioTrackAtPositionZero_whenExtractingTrackIndexes_thenReturnsCorrectMap() {
+        given(simpleExoPlayer.getRendererType(0)).willReturn(C.TRACK_TYPE_AUDIO);
+        Map<TrackType, Integer> expectedMap = new EnumMap<>(TrackType.class);
+        expectedMap.put(TrackType.AUDIO, 0);
```
Mecharyry: You can extract these three lines into `givenTrackPositionZero(TrackType)`

core/src/test/java/com/novoda/noplayer/exoplayer/RendererTrackIndexExtractorTest.java
```diff
@@ -0,0 +1,88 @@
...
+        Map<TrackType, Integer> expectedMap = new EnumMap<>(TrackType.class);
+        expectedMap.put(TrackType.VIDEO, 0);
+
+        Map<TrackType, Integer> trackTypeIntegerMap = extractor.extractFrom(trackCounter, simpleExoPlayer);
+
+        assertThat(trackTypeIntegerMap).isEqualTo(expectedMap);
+    }
+
+    @Test
+    public void givenSubtitlesTrackAtPositionZero_whenExtractingTrackIndexes_thenReturnsCorrectMap() {
```
Mecharyry: I think we can improve the test naming here. I don't know immediately what a `CorrectMap` is. It's used for three tests so do they all have the same "Correct Map"?

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/ExoPlayerMappedTrackInfo.java
zegnus: There are lots of classes in the project that are public because are inside packages and they should not be. We will have to flatten all this classes and only expose what's needed. Not int his PR though.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
zegnus: Lots of public methods in this class call `exoPlayer` without checking if the instance is null or not. That should be protected. I've removed the check here but we shouldn't let a null pointer exception to blow up the internals of a public library. We have to review this class.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
```diff
@@ -104,18 +115,55 @@ public void loadVideo(Uri uri, ContentType contentType, ExoPlayerForwarder forwa
...
+    public void clearSubtitleTrack() {
+        subtitleTrackSelector.clearSubtitleTrack(rendererTypeRequester);
+    }
+
+    public void selectFirstAvailableSubtitlesTrack() {
+        subtitleTrackSelector.selectFirstTextTrack(rendererTypeRequester);
+    }
+
+    @VisibleForTesting
+    final RendererTypeRequester rendererTypeRequester = new RendererTypeRequester() {
```
Mecharyry: ✖️  this is a code smell, we should be able to do this without introducing the annotation.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerCreator.java
```diff
@@ -18,6 +19,7 @@
         this.trackSelector = trackSelector;
     }
 
+    @NonNull
```
Mecharyry: This is redundant in my opinion. We should assume that everything is non-null unless told otherwise.

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacade.java
zegnus: This is not an exception, the code can continue and doesn't need to recover. There are no available subtitle tracks in the MediaPlayer implementation. I might agree with @Dorvaryn rising the level of the log to an error though.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
Mecharyry: You can create a `mock` of the `ExoPlayer` which means you can control the call to `exoPlayer.getRendererType(index)`.

core/src/main/java/com/novoda/noplayer/exoplayer/RendererTypeRequester.java
```diff
@@ -0,0 +1,6 @@
+package com.novoda.noplayer.exoplayer;
+
+public interface RendererTypeRequester {
```
Mecharyry: 💯

core/src/main/java/com/novoda/noplayer/PlayerSubtitleTrack.java
zegnus: Cannot do that right now because its being used inside the package `exoplayer`. We have to flatten this project properly https://github.com/novoda/no-player/pull/47#discussion_r122642440

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
```diff
@@ -140,26 +140,30 @@ private void setExoPlayerTextOutput(TextRenderer.Output output) {
...
+        subtitleTrackSelector.clearSubtitleTrack(rendererTypeRequester);
     }
 
     public void selectFirstAvailableSubtitlesTrack() {
-        subtitleTrackSelector.selectFirstTextTrack();
+        subtitleTrackSelector.selectFirstTextTrack(rendererTypeRequester);
     }
+
+    @VisibleForTesting
+    final RendererTypeRequester rendererTypeRequester = new RendererTypeRequester() {
```
Mecharyry: This is used by the `rendererTrackIndexExtractor`but we are passing this all the way down this way. Is it possible to change how the `rendererTrackIndexExtractor` is created to make this a more solid dependency?

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImplTest.java
zegnus: IDE complains `expression expected`

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/RendererTrackIndexExtractor.java
```diff
@@ -0,0 +1,21 @@
...
+            int rendererType = rendererTypeRequester.getRendererTypeFor(i);
+
+            if ((trackType == TrackType.AUDIO && rendererType == C.TRACK_TYPE_AUDIO)
+                    || (trackType == TrackType.VIDEO && rendererType == C.TRACK_TYPE_VIDEO)
+                    || (trackType == TrackType.TEXT && rendererType == C.TRACK_TYPE_TEXT)) {
+                return i;
+            }
+        }
+
+        return -1;
```
Mecharyry: Is it safe to return `-1`?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
zegnus: Happy to pair with you finding a better solution

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerCreator.java
zegnus: Sure, but the IDE doesn't assume it rising a null pointer problem

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerFacade.java
zegnus: We need to call the native `exoPlayer.getRendererType` that is instantiated after its creation, we do not have it in the factory. Happy to pair with you on this to find a better solution.

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/RendererTrackIndexExtractor.java
zegnus: technically it is, won't crash. What it means is that you are trying to select a track that is not supported. But that would be really weird because you have to pass a track that you get from the player itself...

Mecharyry: I'm asking because this `int trackCount = trackGroups[rendererIndex].get(groupIndex).length;` which is in the `getAdaptiveSupport`. So `-1` may not work in every case that this is returned.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerCreator.java
Mecharyry: Ah I see, fair enough 😄

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImplTest.java
Mecharyry: That's not what I meant. Switch `Object` for `Void`.

core/src/main/java/com/novoda/noplayer/PlayerSubtitleTrack.java
Dorvaryn: 👍 as long as we track it so that we can address that before release.

core/src/main/java/com/novoda/noplayer/PlayerView.java
```diff
@@ -15,4 +19,6 @@
     void showSubtitles();
 
     void hideSubtitles();
+
+    void setSubtitleCue(List<Cue> cues);
```
Dorvaryn: It would be better if we could avoid exposing the Cues here as it is an exoPlayer class. Maybe we could create a Cue class of our own ?

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -188,6 +196,7 @@ public void detach(PlayerView playerView) {
         surfaceHolderRequester = null;
         listenersHolder.removeStateChangedListener(playerView.getStateChangedListener());
         listenersHolder.removeVideoSizeChangedListener(playerView.getVideoSizeChangedListener());
+        this.playerView = null;
```
Dorvaryn: We should unset the `SubtitleRendererOutput` here as it references the playerView.

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -196,11 +205,45 @@ public void selectAudioTrack(PlayerAudioTrack audioTrack) {
...
+    }
+
+    @Override
+    public void showFirstAvailableSubtitleTrack() {
+        setSubtitleRenderer();
+        exoPlayer.selectFirstAvailableSubtitlesTrack();
+        playerView.showSubtitles();
+    }
+
+    private void setSubtitleRenderer() {
```
Dorvaryn: `setSubtitleRendererOutput`

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -196,11 +205,45 @@ public void selectAudioTrack(PlayerAudioTrack audioTrack) {
...
+        exoPlayer.setSubtitleRendererOutput(new TextRenderer.Output() {
+            @Override
+            public void onCues(List<Cue> list) {
+                playerView.setSubtitleCue(list);
+            }
+        });
+    }
+
+    @Override
+    public void hideSubtitleTrack() {
```
Dorvaryn: We should probably also unset the `SubtitleRendererOutput` as it is set in `showSubtitleTrack`

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/RendererTrackIndexExtractor.java
Dorvaryn: `But that would be really weird because you have to pass a track that you get from the player itself...` Maybe this means we should crash here ? This would be a bug/developper error right ?

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -264,11 +266,31 @@ public void selectAudioTrack(PlayerAudioTrack audioTrack) {
...
+    }
+
+    @Override
     public List<PlayerAudioTrack> getAudioTracks() {
         return mediaPlayer.getAudioTracks();
     }
 
     @Override
+    public List<PlayerSubtitleTrack> getSubtitleTracks() {
+        return mediaPlayer.getSubtitleTracks();
```
Dorvaryn: Why does this one delegate to a non implementation when the other methods are handled here ?

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -264,11 +266,31 @@ public void selectAudioTrack(PlayerAudioTrack audioTrack) {
     }
 
     @Override
+    public void showSubtitleTrack(PlayerSubtitleTrack subtitleTrack) {
+        NoPlayerLog.w("Subtitles not implemented for Android Media Player");
```
Dorvaryn: Should we be more aggressive and crash here ? Not sure

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/ExoPlayerMappedTrackInfo.java
juankysoriano: I am asking about this one hehe, Does this one needs to be public? and if so, why?

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -264,11 +267,30 @@ public void selectAudioTrack(PlayerAudioTrack audioTrack) {
...
+
+    @Override
     public List<PlayerAudioTrack> getAudioTracks() {
         return mediaPlayer.getAudioTracks();
     }
 
     @Override
+    public List<PlayerSubtitleTrack> getSubtitleTracks() {
+        SubtitlesError subtitlesError = new SubtitlesError("Subtitles not implemented for Android Media Player", new IllegalStateException());
+        listenersHolder.getErrorListeners().onError(this, subtitlesError);
```
Dorvaryn: 👍

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/ExoPlayerTrackSelector.java
```diff
@@ -4,36 +4,66 @@
...
     }
 
-    void setSelectionOverride(TrackGroupArray trackGroups, MappingTrackSelector.SelectionOverride selectionOverride) {
-        trackSelector.setSelectionOverride(TRACK_TYPE_AUDIO, trackGroups, selectionOverride);
+    void setSelectionOverride(TrackType trackType,
+                              RendererTypeRequester rendererTypeRequester,
+                              TrackGroupArray trackGroups,
+                              MappingTrackSelector.SelectionOverride selectionOverride) {
+        Optional<Integer> audioRendererIndex = rendererTrackIndexExtractor.extract(trackType, mappedTrackInfoLength(), rendererTypeRequester);
+        if (audioRendererIndex.isPresent()) {
```
Dorvaryn: This being absent is a case that should not happen right? I think w should warn the caller it didn't work so that we can bubble up until we can send an error though the ErrorListener.

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/ExoPlayerSubtitleTrackSelector.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public void clearSubtitleTrack(RendererTypeRequester rendererTypeRequester) {
+        trackSelector.clearSelectionOverrideFor(TEXT, rendererTypeRequester);
+    }
+
+    public void selectFirstTextTrack(RendererTypeRequester rendererTypeRequester) {
+        List<PlayerSubtitleTrack> subtitleTracks = getSubtitleTracks(rendererTypeRequester);
+        if (subtitleTracks.isEmpty()) {
+            NoPlayerLog.e("No subtitles tracks available");
```
Dorvaryn: Should we bubble back up to warn through the ErrorListener ? Seems like a feedback the user of the lib should have

zegnus: This method is not used anymore, will remove :)

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/ExoPlayerTrackSelector.java
zegnus: We considered to return a boolean in the method indicating if it was ok or not and bubble up to the client, what about that? This should not be possible as this method requires a trackGroups that is retrieved by a get by type.

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/ExoPlayerMappedTrackInfo.java
zegnus: Was lingering in a test, will fix

core/src/main/java/com/novoda/noplayer/exoplayer/mediasource/ExoPlayerTrackSelector.java
Dorvaryn: It might be good to do yep, if this happens we did something wrong and being warned rather than swallowing would help to catch it faster.

zegnus: done, also added lots of tests https://github.com/novoda/no-player/pull/47/commits/6794007afd1dffbb8fbdb692a7eca9b3c7853aba

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -13,14 +13,17 @@
 import com.novoda.noplayer.PlayerAudioTrack;
 import com.novoda.noplayer.PlayerListenersHolder;
 import com.novoda.noplayer.PlayerState;
+import com.novoda.noplayer.PlayerSubtitleTrack;
 import com.novoda.noplayer.PlayerView;
 import com.novoda.noplayer.SurfaceHolderRequester;
 import com.novoda.noplayer.Timeout;
 import com.novoda.noplayer.VideoDuration;
 import com.novoda.noplayer.VideoPosition;
+import com.novoda.noplayer.exoplayer.playererror.SubtitlesError;
```
juankysoriano: ⚠️ This cannot be here

Mecharyry: Oh 😆

zegnus:
test this please

PR #50, opened 2017-06-23
ALL-3936/Downloaded DRM

Mecharyry:
## Problem
We've added capability to play a DRM protected streaming video, now we need to do the same for a Downloaded one!

## Solution
- `LocalDrmSessionManager` is responsible for acquiring and releasing `DRM` sessions.
- We are representing the states of the `LocalDrmSessionManager` through two `DrmSession`s which can be either `LocalDrmSession` (valid) or `InvalidDrmSession` (invalid/error).
- `DrmSessionStates`are either valid or not, we aren't handling lots of states.
- We wrapped some `byte[]` representing `SessionId`s and `KeySetId` for type safety.

### Test(s) added 
Not just yet. We would need to wrap further the `ExoPlayer` specific classes to do this. We want to have something working and then we will go back and write these. Sorry! 

### Screenshots
No UI changes. We can't test downloads directly on `no-player`, we can't find a local asset that includes the keys 😢 

### Paired with 
@ouchadam

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
@@ -1,0 +1,24 @@
+package com.novoda.noplayer.drm;
+
+import com.google.android.exoplayer2.drm.DrmSessionManager;
+import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
+
+import java.util.UUID;
+
+public class DownloadDrmSessionCreator implements DrmSessionCreator {
+
+    private static final UUID WIDEVINE_MODULAR_UUID = new UUID(0xEDEF8BA979D64ACEL, 0xA3C827DCD51D21EDL);
+
+    private final DownloadedModularDrm downloadedModularDrm;
+    private final FrameworkMediaDrmCreator mediaDrmCreator;
+
+    public DownloadDrmSessionCreator(DownloadedModularDrm downloadedModularDrm, FrameworkMediaDrmCreator mediaDrmCreator) {
+        this.downloadedModularDrm = downloadedModularDrm;
+        this.mediaDrmCreator = mediaDrmCreator;
+    }
+
+    @Override
+    public DrmSessionManager<FrameworkMediaCrypto> create() {
+        return new LocalDrmSessionManager(downloadedModularDrm.getKeySetId(), mediaDrmCreator.create(WIDEVINE_MODULAR_UUID));
+    }
+}
--- a/core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
@@ -5,14 +5,14 @@ import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
 
 import java.util.UUID;
 
-public class DownloadDrmSessionCreator implements DrmSessionCreator {
+class DownloadDrmSessionCreator implements DrmSessionCreator {
 
     private static final UUID WIDEVINE_MODULAR_UUID = new UUID(0xEDEF8BA979D64ACEL, 0xA3C827DCD51D21EDL);
 
     private final DownloadedModularDrm downloadedModularDrm;
     private final FrameworkMediaDrmCreator mediaDrmCreator;
 
-    public DownloadDrmSessionCreator(DownloadedModularDrm downloadedModularDrm, FrameworkMediaDrmCreator mediaDrmCreator) {
+    DownloadDrmSessionCreator(DownloadedModularDrm downloadedModularDrm, FrameworkMediaDrmCreator mediaDrmCreator) {
         this.downloadedModularDrm = downloadedModularDrm;
         this.mediaDrmCreator = mediaDrmCreator;
     }
--- a/core/src/main/java/com/novoda/noplayer/drm/DownloadedModularDrm.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/DownloadedModularDrm.java
@@ -2,5 +2,5 @@ package com.novoda.noplayer.drm;
 
 public interface DownloadedModularDrm extends DrmHandler {
 
-    byte[] getKeySetId();
+    KeySetId getKeySetId();
 }
--- a/core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
@@ -18,7 +18,7 @@ public class DrmSessionCreatorFactory {
                 );
                 return StreamingDrmSessionCreator.newInstance(mediaDrmCallback, new FrameworkMediaDrmCreator());
             case WIDEVINE_MODULAR_DOWNLOAD:
-                return new NoDrmSessionCreator();
+                return new DownloadDrmSessionCreator((DownloadedModularDrm) drmHandler, new FrameworkMediaDrmCreator());
             default:
                 throw PlayerFactory.UnableToCreatePlayerException.noDrmHandlerFor(drmType);
         }
--- a/core/src/main/java/com/novoda/noplayer/drm/InvalidDrmSession.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/InvalidDrmSession.java
@@ -1,5 +1,7 @@
 package com.novoda.noplayer.drm;
 
+import android.support.annotation.Nullable;
+
 import com.google.android.exoplayer2.drm.DrmSession;
 import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
 
@@ -40,6 +42,7 @@ class InvalidDrmSession implements FrameworkDrmSession {
         throw new IllegalStateException();
     }
 
+    @Nullable
     @Override
     public byte[] getOfflineLicenseKeySetId() {
         return ABSENT_OFFLINE_LICENSE_KEY_SET_ID;
--- a/core/src/main/java/com/novoda/noplayer/drm/InvalidDrmSession.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/InvalidDrmSession.java
@@ -44,7 +44,7 @@ class InvalidDrmSession implements FrameworkDrmSession {
     }
 
     @Override
-    public byte[] getSessionId() {
-        return new byte[0];
+    public SessionId getSessionId() {
+        return SessionId.absent();
     }
 }
--- a/core/src/main/java/com/novoda/noplayer/drm/InvalidDrmSession.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/InvalidDrmSession.java
@@ -1,0 +1,50 @@
+package com.novoda.noplayer.drm;
+
+import com.google.android.exoplayer2.drm.DrmSession;
+import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
+
+import java.util.Map;
+
+class InvalidDrmSession implements FrameworkDrmSession {
+
+    private final DrmSessionException drmSessionException;
+
+    InvalidDrmSession(DrmSessionException drmSessionException) {
+        this.drmSessionException = drmSessionException;
+    }
+
+    @Override
+    public int getState() {
+        return DrmSession.STATE_ERROR;
+    }
+
+    @Override
+    public FrameworkMediaCrypto getMediaCrypto() {
+        throw new IllegalStateException();
+    }
+
+    @Override
+    public boolean requiresSecureDecoderComponent(String mimeType) {
+        throw new IllegalStateException();
+    }
+
+    @Override
+    public DrmSessionException getError() {
+        return drmSessionException;
+    }
+
+    @Override
+    public Map<String, String> queryKeyStatus() {
+        throw new IllegalStateException();
+    }
+
+    @Override
+    public byte[] getOfflineLicenseKeySetId() {
+        return null;
+    }
+
+    @Override
+    public byte[] getSessionId() {
+        return new byte[0];
+    }
+}
--- a/core/src/main/java/com/novoda/noplayer/drm/InvalidDrmSession.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/InvalidDrmSession.java
@@ -7,6 +7,8 @@ import java.util.Map;
 
 class InvalidDrmSession implements FrameworkDrmSession {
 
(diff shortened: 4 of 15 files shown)


core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
```diff
@@ -0,0 +1,24 @@
+package com.novoda.noplayer.drm;
+
+import com.google.android.exoplayer2.drm.DrmSessionManager;
+import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
+
+import java.util.UUID;
+
+public class DownloadDrmSessionCreator implements DrmSessionCreator {
+
+    private static final UUID WIDEVINE_MODULAR_UUID = new UUID(0xEDEF8BA979D64ACEL, 0xA3C827DCD51D21EDL);
```
ouchadam: these duplicated constants are all replaced with a single one inside  `DrmSessionCreator` in the next PR

core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
```diff
@@ -0,0 +1,24 @@
+package com.novoda.noplayer.drm;
+
+import com.google.android.exoplayer2.drm.DrmSessionManager;
```
jackSzm: Is there a way to get away with exoplayer2 dependencies in all-4 ?

core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
```diff
@@ -18,7 +18,7 @@ public DrmSessionCreator createFor(DrmType drmType, DrmHandler drmHandler) {
                 );
                 return StreamingDrmSessionCreator.newInstance(mediaDrmCallback, new FrameworkMediaDrmCreator());
             case WIDEVINE_MODULAR_DOWNLOAD:
-                return new NoDrmSessionCreator();
+                return new DownloadDrmSessionCreator((DownloadedModularDrm) drmHandler, new FrameworkMediaDrmCreator());
```
jackSzm: Why do we need to cast here ? Is it safe ? Maybe we could avoid it 🤔

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSessionManager.java
```diff
@@ -0,0 +1,55 @@
...
+
+    LocalDrmSessionManager(KeySetId keySetIdToRestore, ExoMediaDrm<FrameworkMediaCrypto> mediaDrm) {
+        this.keySetIdToRestore = keySetIdToRestore;
+        this.mediaDrm = mediaDrm;
+    }
+
+    @TargetApi(Build.VERSION_CODES.KITKAT)
+    @Override
+    public DrmSession<FrameworkMediaCrypto> acquireSession(Looper playbackLooper, DrmInitData drmInitData) {
+        DrmSession<FrameworkMediaCrypto> drmSession;
```
jackSzm: Instead of creating this variable and returning it after `catch` can't we just directly return in both `try` and `catch` ?

core/src/main/java/com/novoda/noplayer/drm/DownloadedModularDrm.java
```diff
@@ -2,5 +2,5 @@
 
 public interface DownloadedModularDrm extends DrmHandler {
 
-    byte[] getKeySetId();
+    KeySetId getKeySetId();
```
joetimmins: ❤️

core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
```diff
@@ -0,0 +1,24 @@
...
+import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
+
+import java.util.UUID;
+
+public class DownloadDrmSessionCreator implements DrmSessionCreator {
```
joetimmins: This can be a package-private class - only the interface should be public here really

core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
```diff
@@ -0,0 +1,24 @@
+package com.novoda.noplayer.drm;
```
jackSzm: Since all this is ExoPlayer (Not AndroidMediaPlayer) specific , can we move it into exoplayer package ?

jackSzm: So all the exoplayer2 dependencies are in exoplayer.

core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
ouchadam: it's for sugar 

`playerFactory.create(drmType, drmHandler)`

rather than 
```
playerFactory.createModularStreamingPlayer(modularStreamingHandler)
playerFactory.createModularDownloadPlayer(modularDownloadHandler)
playerFactory.createClassicPlayer(classicType)
playerFactory.createNoDrmPlayer(noDrmType)
```

joetimmins: `CreatorFactory` is a bit of a `FactoryFactory` type thing isn't it? Can we not just make a class that returns the correct `DrmSessionManager` implementation directly?

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSessionManager.java
```diff
@@ -0,0 +1,55 @@
...
+
+        try {
+            SessionId sessionId = SessionId.of(mediaDrm.openSession());
+            FrameworkMediaCrypto mediaCrypto = mediaDrm.createMediaCrypto(WIDEVINE_MODULAR_UUID, sessionId.asBytes());
+
+            mediaDrm.restoreKeys(sessionId.asBytes(), keySetIdToRestore.asBytes());
+
+            drmSession = new LocalDrmSession(mediaCrypto, keySetIdToRestore, sessionId);
+        } catch (NotProvisionedException | MediaCryptoException | ResourceBusyException e) {
+            drmSession = new InvalidDrmSession(new DrmSession.DrmSessionException(e));
```
joetimmins: Do we need to log the exception here at all? Or rethrow a `DrmSessionManagerCreationException`? (that exception is never used, btw)

core/src/main/java/com/novoda/noplayer/drm/SessionId.java
```diff
@@ -0,0 +1,50 @@
+package com.novoda.noplayer.drm;
+
+import java.util.Arrays;
+
+class SessionId {
+
+    private final byte[] sessionId;
+
+    public static SessionId absent() {
```
joetimmins: ❤️ no primitive obsession here

core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
Mecharyry: We were planning on going back through the project when the work of migrating is done to flatten the package structure and make it harder to get at some of the internal dependencies.

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSessionManager.java
Mecharyry: This is actually tackled in the next PR, we forward all errors from the DRM part to the expected listeners on the `NoPlayer` side.

core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
```diff
@@ -18,7 +18,7 @@ public DrmSessionCreator createFor(DrmType drmType, DrmHandler drmHandler) {
                 );
                 return StreamingDrmSessionCreator.newInstance(mediaDrmCallback, new FrameworkMediaDrmCreator());
             case WIDEVINE_MODULAR_DOWNLOAD:
-                return new NoDrmSessionCreator();
```
joetimmins: This guy can be package-private too

core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
```diff
@@ -18,7 +18,7 @@ public DrmSessionCreator createFor(DrmType drmType, DrmHandler drmHandler) {
                 );
                 return StreamingDrmSessionCreator.newInstance(mediaDrmCallback, new FrameworkMediaDrmCreator());
```
joetimmins: This guy can be / should be package-private too

core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
ouchadam: it's because the lifecycle of the player is 

```
create a player -> for content and drm type and then attach listeners
load a video -> create and start drm session passing in previously set up listeners
```

joetimmins: Ah okay. Horrible, but fine 👍

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSessionManager.java
joetimmins: Ok nice one. Just checking

core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
ouchadam: I think the names aren't particular helpful at the moment, the plan is to try and flatten a bunch of the creators/factories, hopefully that'll help a bunch

Mecharyry:
@joetimmins I've made sure that the classes are now package scope. @jackSzm I think we have addressed your comments too?

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSession.java
```diff
@@ -0,0 +1,57 @@
...
+        return mediaCrypto;
+    }
+
+    @Override
+    public boolean requiresSecureDecoderComponent(String mimeType) {
+        return mediaCrypto.requiresSecureDecoderComponent(mimeType);
+    }
+
+    @Override
+    public DrmSessionException getError() {
```
zegnus: `@Nullable`

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSessionManager.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private final KeySetId keySetIdToRestore;
+    private final ExoMediaDrm<FrameworkMediaCrypto> mediaDrm;
+
+    LocalDrmSessionManager(KeySetId keySetIdToRestore, ExoMediaDrm<FrameworkMediaCrypto> mediaDrm) {
+        this.keySetIdToRestore = keySetIdToRestore;
+        this.mediaDrm = mediaDrm;
+    }
+
+    @TargetApi(Build.VERSION_CODES.KITKAT)
```
zegnus: Check that this code doesn't actually crash on a device < KIT_KAT please, confirm before merging ;)

core/src/main/java/com/novoda/noplayer/drm/SessionId.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public static SessionId of(byte[] sessionId) {
+        return new SessionId(Arrays.copyOf(sessionId, sessionId.length));
+    }
+
+    private SessionId(byte[] sessionId) {
+        this.sessionId = sessionId;
+    }
+
+    public byte[] asBytes() {
```
zegnus: 👍

core/src/main/java/com/novoda/noplayer/drm/InvalidDrmSession.java
```diff
@@ -0,0 +1,52 @@
...
+        return drmSessionException;
+    }
+
+    @Override
+    public Map<String, String> queryKeyStatus() {
+        throw new IllegalStateException();
+    }
+
+    @Override
+    public byte[] getOfflineLicenseKeySetId() {
```
zegnus: `@nullable`

ouchadam: this isn't our interface, not sure if it'll make a difference

zegnus: do you mean that we do not call this method ourselves?

ouchadam: yep, it's used internally be exoplayer

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSessionManager.java
ouchadam: will do 👍  (I would argue it isn't the responsibility of the library to handle the case where a client has tried to load/play modular content on a device that doesn't support mod)

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSessionManager.java
Mecharyry: As per Slack -> "Something something don't use for control flow something reason something" 😆

Mecharyry: we think this is the reasoning 😄  http://wiki.c2.com/?DontUseExceptionsForFlowControl

core/src/main/java/com/novoda/noplayer/drm/InvalidDrmSession.java
zegnus: cool 👍 no need then

zegnus:
added do not merge just to wait for https://github.com/novoda/no-player/pull/50#discussion_r123739875 response

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSessionManager.java
ouchadam: yep, it all works as long as you don't try and play content the device doesn't support

zegnus: We'll try to notify the user that cannot use the library methods below 4.4 in another PR 👍

PR #51, opened 2017-06-26
ALL-3936/Error listeners

Mecharyry:
## Problem
We created a `LocalDrmSession` but we ignored bubbling any errors up to our `no-player` error listeners. 

## Solution
Create a `DrmSessionErrorForwarder` to transform and forward our `ExoPlayer` DRM error into a `no-player` specific `Player.Error`.

We had `UUID`in three or four places for `Widevine`, we have now exposed it in a common interface. 

### Test(s) added 
No, we would need to wrap a lot of `ExoPlayer` classes in order to test the DRM part. We will do this at the end.

### Screenshots
No UI changes.

### Paired with 
@ouchadam

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
@@ -16,6 +16,6 @@ class DownloadDrmSessionCreator implements DrmSessionCreator {
 
     @Override
     public DrmSessionManager<FrameworkMediaCrypto> create(DefaultDrmSessionManager.EventListener eventListener) {
-        return new LocalDrmSessionManager(downloadedModularDrm.getKeySetId(), mediaDrmCreator.create(WIDEVINE_MODULAR_UUID), eventListener, WIDEVINE_MODULAR_UUID);
+        return LocalDrmSessionManager.newInstance(downloadedModularDrm.getKeySetId(), mediaDrmCreator.create(WIDEVINE_MODULAR_UUID), eventListener, WIDEVINE_MODULAR_UUID);
     }
 }
--- a/core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
@@ -20,6 +20,6 @@ class DownloadDrmSessionCreator implements DrmSessionCreator {
 
     @Override
     public DrmSessionManager<FrameworkMediaCrypto> create(DefaultDrmSessionManager.EventListener eventListener) {
-        return new LocalDrmSessionManager(downloadedModularDrm.getKeySetId(), mediaDrmCreator.create(WIDEVINE_MODULAR_UUID));
+        return new LocalDrmSessionManager(downloadedModularDrm.getKeySetId(), mediaDrmCreator.create(WIDEVINE_MODULAR_UUID), eventListener);
     }
 }
--- a/core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
+++ b/core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
@@ -4,11 +4,7 @@ import com.google.android.exoplayer2.drm.DefaultDrmSessionManager;
 import com.google.android.exoplayer2.drm.DrmSessionManager;
 import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
 
-import java.util.UUID;
-
 class DownloadDrmSessionCreator implements DrmSessionCreator {
-
-    private static final UUID WIDEVINE_MODULAR_UUID = new UUID(0xEDEF8BA979D64ACEL, 0xA3C827DCD51D21EDL);
 
     private final DownloadedModularDrm downloadedModularDrm;
     private final FrameworkMediaDrmCreator mediaDrmCreator;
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/novoda/noplayer/drm/DownloadDrmSessionCreator.java
```diff
@@ -1,14 +1,11 @@
...
 
+import com.google.android.exoplayer2.drm.DefaultDrmSessionManager;
 import com.google.android.exoplayer2.drm.DrmSessionManager;
 import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
 
-import java.util.UUID;
-
 class DownloadDrmSessionCreator implements DrmSessionCreator {
 
-    private static final UUID WIDEVINE_MODULAR_UUID = new UUID(0xEDEF8BA979D64ACEL, 0xA3C827DCD51D21EDL);
```
zegnus: 👍

core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreator.java
```diff
@@ -2,13 +2,18 @@
...
 
+import com.google.android.exoplayer2.drm.DefaultDrmSessionManager;
 import com.google.android.exoplayer2.drm.DrmSessionManager;
 import com.google.android.exoplayer2.drm.FrameworkMediaCrypto;
 
+import java.util.UUID;
+
 public interface DrmSessionCreator {
 
+    UUID WIDEVINE_MODULAR_UUID = new UUID(0xEDEF8BA979D64ACEL, 0xA3C827DCD51D21EDL);
```
juankysoriano: 💯

core/src/main/java/com/novoda/noplayer/exoplayer/forwarder/ExoPlayerDrmSessionEventListener.java
```diff
@@ -0,0 +1,43 @@
+package com.novoda.noplayer.exoplayer.forwarder;
+
+import com.google.android.exoplayer2.drm.DefaultDrmSessionManager;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class ExoPlayerDrmSessionEventListener implements DefaultDrmSessionManager.EventListener {
```
juankysoriano: I think we were naming classes like this as `Listeners` ? but I am not sure if that was for the class or for the field when using it

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSessionManager.java
```diff
@@ -17,14 +19,27 @@
...
 
-    private static final UUID WIDEVINE_MODULAR_UUID = new UUID(0xEDEF8BA979D64ACEL, 0xA3C827DCD51D21EDL);
-
     private final KeySetId keySetIdToRestore;
     private final ExoMediaDrm<FrameworkMediaCrypto> mediaDrm;
+    private final DefaultDrmSessionManager.EventListener eventListener;
+    private final UUID drmScheme;
+    private final Handler handler;
+
+    static LocalDrmSessionManager newInstance(KeySetId keySetIdToRestore, ExoMediaDrm<FrameworkMediaCrypto> mediaDrm, DefaultDrmSessionManager.EventListener eventListener, UUID drmScheme) {
```
juankysoriano: I think the `UUID` parameter would feel better next to the `ExoMediaDrm` and leaving the listener for the last. Same for the constructor. 

Of course this is not a blocker

core/src/main/java/com/novoda/noplayer/exoplayer/forwarder/DrmSessionErrorForwarder.java
```diff
@@ -0,0 +1,38 @@
...
+    private final ErrorListeners errorListeners;
+
+    DrmSessionErrorForwarder(Player player, ErrorListeners errorListeners) {
+        this.player = player;
+        this.errorListeners = errorListeners;
+    }
+
+    @Override
+    public void onDrmKeysLoaded() {
+        // TODO: Are we interested?
```
juankysoriano: @ouchadam what would be nice use case of this not-used callbacks? 

If there are none I would suggest replacing the TODO with a `not-used` comment

ouchadam: currently all the `*forwarders` have a todo mentioning if we should track them, would prefer to do them all at the same time

core/src/main/java/com/novoda/noplayer/exoplayer/forwarder/ExoPlayerDrmSessionEventListener.java
Mecharyry: These are the internal listeners that are attached to either `exo-player` or `media-player`. The ones you are thinking of are the ones exposed on the public api.

core/src/main/java/com/novoda/noplayer/exoplayer/forwarder/DrmSessionErrorForwarder.java
juankysoriano: 👍

core/src/main/java/com/novoda/noplayer/exoplayer/forwarder/ExoPlayerDrmSessionEventListener.java
juankysoriano: 👍

core/src/main/java/com/novoda/noplayer/drm/LocalDrmSessionManager.java
ouchadam: 7e59b0e

PR #52, opened 2017-06-26
ALL-3936/Intel hack

Mecharyry:
## Problem
There's a bug in Intel firmware that means that secure decoders fail to decode content that does not require a secure decoder. e.g. Widevine modular L3.

## Solution
We've added the ability to override the secure decoder selection from the client. 

We copied the `DefaultRenderersFactory` so that we could supply a custom `CodecSelector` to the audio and video renderers. 

### Test(s) added 
YES! we added tests to the `CodecSelector` to ensure our logic was correct.

### Screenshots
No UI changes.

### Paired with 
@ouchadam

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/exoplayer/CodecSelector.java
+++ b/core/src/main/java/com/novoda/noplayer/exoplayer/CodecSelector.java
@@ -1,0 +1,25 @@
+package com.novoda.noplayer.exoplayer;
+
+import com.google.android.exoplayer2.mediacodec.MediaCodecInfo;
+import com.google.android.exoplayer2.mediacodec.MediaCodecSelector;
+import com.google.android.exoplayer2.mediacodec.MediaCodecUtil;
+
+class CodecSelector implements MediaCodecSelector {
+
+    private final boolean useSecureCodec;
+
+    CodecSelector(boolean useSecureCodec) {
+        this.useSecureCodec = useSecureCodec;
+    }
+
+    @Override
+    public MediaCodecInfo getDecoderInfo(String mimeType, boolean requiresSecureDecoder)
+            throws MediaCodecUtil.DecoderQueryException {
+        return MediaCodecUtil.getDecoderInfo(mimeType, useSecureCodec);
+    }
+
+    @Override
+    public MediaCodecInfo getPassthroughDecoderInfo() throws MediaCodecUtil.DecoderQueryException {
+        return MediaCodecUtil.getPassthroughDecoderInfo();
+    }
+}
--- a/core/src/main/java/com/novoda/noplayer/exoplayer/CodecSelector.java
+++ b/core/src/main/java/com/novoda/noplayer/exoplayer/CodecSelector.java
@@ -4,22 +4,44 @@ import com.google.android.exoplayer2.mediacodec.MediaCodecInfo;
 import com.google.android.exoplayer2.mediacodec.MediaCodecSelector;
 import com.google.android.exoplayer2.mediacodec.MediaCodecUtil;
 
+/**
+ * CodecSelector that allows overriding of the requiresSecureDecoder to allow
+ * downgrading of security but not upgrading.
+ */
 class CodecSelector implements MediaCodecSelector {
 
-    private final boolean useSecureCodec;
+    private final InternalMediaCodecUtil internalMediaCodecUtil;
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/novoda/noplayer/exoplayer/CodecSelector.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * CodecSelector that allows overriding of the requiresSecureDecoder to allow
+ * downgrading of security but not upgrading.
+ */
+class CodecSelector implements MediaCodecSelector {
+
+    private final InternalMediaCodecUtil internalMediaCodecUtil;
+    private final boolean overrideSecureDecoder;
+
+    public static CodecSelector newInstance(boolean useSecureCodec) {
```
zegnus: You can create two factories instead ;)

core/src/main/java/com/novoda/noplayer/player/PlayerFactory.java
```diff
@@ -14,6 +14,8 @@
 
 public class PlayerFactory {
 
+    private static final boolean USE_SECURE_CODEC = false;
```
zegnus: Can the user of the library decide?

Mecharyry: Yes, they can. We need this for the `intel` hack. Previously we did the hack in the library but it isn't really it's responsibility to know the hack, it relies on the content security which the library does not know about.

core/src/main/java/com/novoda/noplayer/player/PlayerFactory.java
```diff
@@ -41,21 +43,25 @@ public Player create() {
...
             if (player.supports(drmType)) {
-                return createPlayerForType(player, drmType, drmHandler);
+                return createPlayerForType(player, drmType, drmHandler, useSecureCodec);
             }
         }
         throw UnableToCreatePlayerException.unhandledDrmType(drmType);
     }
 
-    private Player createPlayerForType(PlayerType playerType, DrmType drmType, DrmHandler drmHandler) {
+    private Player createPlayerForType(PlayerType playerType, DrmType drmType, DrmHandler drmHandler, boolean useSecureCodec) {
```
zegnus: what about exposing a single `createExoPlayerWithSecureCoded` instead?

core/src/main/java/com/novoda/noplayer/exoplayer/CodecSelector.java
joetimmins: `useSecureCodec` seems to mean the opposite of `overrideSecureDecoder` no? 

`useSecureCodec = true` -> "go ahead, use the secure codec"
`overrideSecureCodec = true` -> "override that secure codec, use the insecure one instead"

So... which one is it?

joetimmins:
I was trying to think of ways you could avoid having the `boolean` being passed round so many places, but to be honest given you're wrapping a framework call that needs that boolean, the other ways i thought of weren't really much nicer.

core/src/main/java/com/novoda/noplayer/exoplayer/CodecSelector.java
Mecharyry: We got confused with this `CodecSelector` too, we decided to invert the logic and do some renaming to make it make sense. 😄  65008cf

core/src/main/java/com/novoda/noplayer/player/PlayerFactory.java
ouchadam: https://github.com/novoda/no-player/pull/52/files/4748c1dd02defdb565cd7174a38deb90ee214d9d#r124011428

this would no longer be needed as 99% of cases would never need to manually downgrade the secure decoder 🐱

zegnus: I think that its ok to leave internal implementation with a boolean (as is a less big deal) but with the public API creating two different methods is clearer. If its not widely used its even better, you keep the normal method standardised and another with that is fit to purpose with a good naming ;)

zegnus:
There is a non-solved conversation about removing the user-facing boolean parameter, we will merge this and decide about how to tackle it tomorrow morning and amend the code if necessary.

Issue #40, opened 2017-06-07
JavaDoc Public APIs


ouchadam:
#### Problem
We have very little to no documentation for our public APIs.

#### Potential Solution

We should javadoc it, especially considering some method have side effects which will effect the calling of others. Some methods throw exceptions which the clients need to know about and handle.

ouchadam:
fixed by #58

PR #54, opened 2017-06-26
ALL-3936/Assert support DRM

Mecharyry:
## Problem
We were crashing when a device that does not support Media DRM apis was attempting to play content. The error message was not useful to clients as it did not provide information as to why the crash occurred.  

## Solution
Check the device version when trying to create a `Player` with a `DrmHandler`, if it does not meet the required api level then we crash with a meaningful message. 

Changed the scope of some associated classes, removing static factory methods where possible. 

### Test(s) added 
Add tests to the `DrmSessionCreatorFactory` to assert the exceptions and messages.

### Screenshots
No UI changes.

### Paired with 
@ouchadam


core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
```diff
@@ -1,27 +1,66 @@
...
 
+    private final AndroidDeviceVersion androidDeviceVersion;
+    private final Handler handler;
+
+    public DrmSessionCreatorFactory(AndroidDeviceVersion androidDeviceVersion, Handler handler) {
+        this.androidDeviceVersion = androidDeviceVersion;
+        this.handler = handler;
+    }
+
     public DrmSessionCreator createFor(DrmType drmType, DrmHandler drmHandler) {
```
zegnus: It seems that we can throw unchecked exceptions, we can help the user of this method to manage expectations if we document them ;)

core/src/main/java/com/novoda/noplayer/drm/StreamingDrmSessionCreator.java
```diff
@@ -19,12 +18,7 @@
     private final FrameworkMediaDrmCreator frameworkMediaDrmCreator;
     private final Handler handler;
 
-    static StreamingDrmSessionCreator newInstance(MediaDrmCallback mediaDrmCallback, FrameworkMediaDrmCreator frameworkMediaDrmCreator) {
```
zegnus: 👍

core/src/main/java/com/novoda/noplayer/drm/provision/ProvisionExecutor.java
```diff
@@ -12,13 +12,7 @@
     private final HttpPoster httpPoster;
     private final ProvisioningCapabilities capabilities;
 
-    public static ProvisionExecutor newInstance() {
```
zegnus: 👍

core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
ouchadam: do you think we should add a checked exception to the `PlayerBuilder.build()`?

if so we can propagate those errors up into `PlayerCreationExceptions`

core/src/main/java/com/novoda/noplayer/exoplayer/NoPlayerExoPlayerCreator.java
```diff
@@ -0,0 +1,88 @@
...
+import com.novoda.noplayer.drm.DrmSessionCreator;
+import com.novoda.noplayer.exoplayer.forwarder.ExoPlayerForwarder;
+import com.novoda.noplayer.exoplayer.mediasource.ExoPlayerAudioTrackSelector;
+import com.novoda.noplayer.exoplayer.mediasource.ExoPlayerSubtitleTrackSelector;
+import com.novoda.noplayer.exoplayer.mediasource.ExoPlayerTrackSelector;
+import com.novoda.noplayer.exoplayer.mediasource.MediaSourceFactory;
+
+public class NoPlayerExoPlayerCreator {
+
+    private final InternalCreator internalCreator;
```
zegnus: I'm not sure we gain much doing this (apart from creating a whole class). You can just hold a reference to the handler and then use a static method just use a local create method. We then reduce indirection.

core/src/main/java/com/novoda/noplayer/mediaplayer/NoPlayerMediaPlayerCreator.java
```diff
@@ -0,0 +1,52 @@
...
+import com.novoda.noplayer.Heart;
+import com.novoda.noplayer.LoadTimeout;
+import com.novoda.noplayer.Player;
+import com.novoda.noplayer.PlayerListenersHolder;
+import com.novoda.noplayer.SystemClock;
+import com.novoda.noplayer.mediaplayer.forwarder.MediaPlayerForwarder;
+
+public class NoPlayerMediaPlayerCreator {
+
+    private final InternalCreator internalCreator;
```
zegnus: Same as before

core/src/main/java/com/novoda/noplayer/exoplayer/NoPlayerExoPlayerCreator.java
ouchadam: the internal classes are for testing

zegnus: Also if you do this, you can remove the static newInstance and simply use the normal constructor

core/src/main/java/com/novoda/noplayer/player/PlayerFactory.java
```diff
@@ -1,40 +1,49 @@
...
 
-    public PlayerFactory(Context context, PrioritizedPlayerTypes prioritizedPlayerTypes) {
-        this(context, prioritizedPlayerTypes, ExoPlayerCreator.newInstance(), MediaPlayerCreator.newInstance(), new DrmSessionCreatorFactory());
+    public static PlayerFactory newInstance(Context context, PrioritizedPlayerTypes prioritizedPlayerTypes) {
+        Handler handler = new Handler(Looper.getMainLooper());
+        DrmSessionCreatorFactory drmSessionCreatorFactory = new DrmSessionCreatorFactory(AndroidDeviceVersion.newInstance(), handler);
+        return new PlayerFactory(
+                context,
+                prioritizedPlayerTypes,
+                NoPlayerExoPlayerCreator.newInstance(handler),
```
zegnus: If you remove the internal indirection for the creation then you can just use `new NoPlayerE....(handler)`

zegnus:
Please review public facing methods and document all possible exceptions that those methods can throw, also document those exceptions in internal implementation so that developers can keep track of them :)

core/src/main/java/com/novoda/noplayer/drm/DrmSessionCreatorFactory.java
zegnus: Can we recover from those exceptions or is a fatal crash?

ouchadam: nope, they'll be thrown if the device or available players can't handle the drm types (content types should eventually be taken into account here as well)

zegnus: if we cannot recover this is a fatal exception, don't check them then

core/src/main/java/com/novoda/noplayer/exoplayer/NoPlayerExoPlayerCreator.java
zegnus: Checked the code, ok.

`ExoPlayerTwoImpl` and `AndroidMediaPlayerImpl` class can be package protected ;)

Mecharyry: We did this in the #55 PR 😄

zegnus: ok 👍

ouchadam:
@zegnus We'll add javadoc to the `PlayerBuilder` in the lastest PR #40

core/src/main/java/com/novoda/noplayer/drm/provision/HttpPoster.java
```diff
@@ -6,7 +6,7 @@
 import java.net.HttpURLConnection;
 import java.net.URL;
 
-class HttpPoster {
+public class HttpPoster {
```
juankysoriano: Thy is this class now public?

juankysoriano: I would rather keep this `package` and in case make a `newInstance`  on the `ProvisioningExecutor`. I think that's one of the little benefits, on my opinion, of the `newInstance` and therefore I don't see justification from removing it from `ProvisioningExecutor`

core/src/main/java/com/novoda/noplayer/exoplayer/NoPlayerExoPlayerCreator.java
```diff
@@ -0,0 +1,88 @@
...
+        return new NoPlayerExoPlayerCreator(internalCreator);
+    }
+
+    NoPlayerExoPlayerCreator(InternalCreator internalCreator) {
+        this.internalCreator = internalCreator;
+    }
+
+    public Player createExoPlayer(Context context, DrmSessionCreator drmSessionCreator, boolean downgradeSecureDecoder) {
+        ExoPlayerTwoImpl player = internalCreator.create(context, drmSessionCreator, downgradeSecureDecoder);
+        player.initialise();
```
juankysoriano: 👍

core/src/main/java/com/novoda/noplayer/player/PlayerFactory.java
```diff
@@ -58,10 +67,10 @@ public Player create(DrmType drmType, DrmHandler drmHandler, boolean downgradeSe
     private Player createPlayerForType(PlayerType playerType, DrmType drmType, DrmHandler drmHandler, boolean downgradeSecureDecoder) {
         switch (playerType) {
             case MEDIA_PLAYER:
-                return mediaPlayerCreator.createMediaPlayer(context);
+                return noPlayerMediaPlayerCreator.createMediaPlayer(context);
```
juankysoriano: An alternative could have been not having the `InternalCreator` in the creators (therefore you don't need testing) and doing the test for the `player.initialise()` here (which I believe is the bit you wanted to test)

ouchadam: this would mean making all the impls public as they live inside their own player specific packages

core/src/test/java/com/novoda/noplayer/mediaplayer/NoPlayerMediaPlayerCreatorTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.verify;
+
+public class NoPlayerMediaPlayerCreatorTest {
```
juankysoriano: read my comment on `NoPlayerFactory`, we could take rid of two test classes.

core/src/main/java/com/novoda/noplayer/player/PlayerFactory.java
juankysoriano: Well, we can still have the `InternalCreator` it that benefits us for reducing the scope of those classes, but I still believe we can benefit from doing `player.initialise()` here instead of the two creators. 

Not a blocker, just a thought

core/src/main/java/com/novoda/noplayer/drm/provision/HttpPoster.java
ouchadam: these classes were being used in the client project, we'll replace this class with an interface and make this impl package again

core/src/main/java/com/novoda/noplayer/player/PlayerFactory.java
ouchadam: I mentioned this the no-player room but I can say again 👻 

It comes down to which layer should provide a fully instantiated player, should it be the concrete player packages (media/exo) which allows all the classes to remain package local or the noplayer root and then expose implementation details about the separate players

juankysoriano: ahhhhhhhhhhhhhhhhh

I undertand now, thanks. Yes let's keep this like it is

core/src/test/java/com/novoda/noplayer/mediaplayer/NoPlayerMediaPlayerCreatorTest.java
juankysoriano: scratch this

PR #55, opened 2017-06-26
ALL-3936/Package restructure

Mecharyry:
## Problem
Lots and lots of public classes that could be package if they were in the right place.

## Solution
Move classes to logically grouped areas and make them package scope.

### Test(s) added 
No, just updating packages.

### Screenshots
No UI changes.

### Paired with 
@ouchadam


core/src/main/java/com/novoda/noplayer/exoplayer/playererror/ExoPlayerErrorFactory.java
```diff
@@ -1,27 +1,21 @@
...
-import com.novoda.noplayer.exoplayer.playererror.DrmDecryptionError;
-import com.novoda.noplayer.exoplayer.playererror.DrmRequestError;
-import com.novoda.noplayer.exoplayer.playererror.InvalidResponseCodeError;
-import com.novoda.noplayer.exoplayer.playererror.MalformedContentError;
-import com.novoda.noplayer.exoplayer.playererror.UnknownError;
 
 import java.io.IOException;
 
-final class ExoPlayerErrorFactory {
+public final class ExoPlayerErrorFactory {
```
zegnus: Is this ok that is available from the client?

core/src/main/java/com/novoda/noplayer/model/TextCues.java
```diff
@@ -12,7 +12,7 @@ public TextCues(List<Cue> cues) {
         this.cues = cues;
     }
 
-    List<Cue> cues() {
+    public List<Cue> cues() {
```
zegnus: This class needs to be public facing?

core/src/main/java/com/novoda/noplayer/exoplayer/playererror/ExoPlayerErrorFactory.java
Mecharyry: I don't see the problem with this being accessible from the client. They will only get a specific `Player.PlayerError` based on an `Exception`. They can't extend it or anything else.

core/src/main/java/com/novoda/noplayer/model/TextCues.java
Mecharyry: It wouldn't need to be if we collapsed some more packages but all of our models now belong in the `model` package meaning they need to be `public`. Personally, I would find it strange if we didn't expose our `NoPlayer` models.

core/src/main/java/com/novoda/noplayer/exoplayer/playererror/ExoPlayerErrorFactory.java
zegnus: If this is public a client can do `ExoPlayerErrorFactory.errorFor(...)`. Is this part of the public API?

core/src/main/java/com/novoda/noplayer/model/TextCues.java
zegnus: If the client needs to have those models then ok, but if it doesn't then these models become a contract with the clients

core/src/main/java/com/novoda/noplayer/exoplayer/playererror/ExoPlayerErrorFactory.java
Mecharyry: We opted to rename as `Mapper` and move to the package where it is used 6b400cf

core/src/main/java/com/novoda/noplayer/model/TextCues.java
ouchadam: `TextCues` is part of the `NoPlayer` contract as it's used in `PlayerView`

zegnus: ok

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerFacadeTest.java
```diff
@@ -391,7 +391,7 @@ public void whenGettingAudioTracks_thenDelegatesToTrackSelector() {
         @Mock
         DrmSessionCreator drmSessionCreator;
         @Mock
-        ExoPlayerDrmSessionEventListener drmSessionEventListener;
+        DefaultDrmSessionManager.EventListener drmSessionEventListener;
```
joetimmins: static import?

core/src/main/java/com/novoda/noplayer/drm/provision/ProvisionExecutorImpl.java
```diff
@@ -0,0 +1,36 @@
+package com.novoda.noplayer.drm.provision;
+
+import com.novoda.noplayer.drm.ModularDrmProvisionRequest;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+
+class ProvisionExecutorImpl implements ProvisionExecutor {
```
joetimmins: Can we call this something that doesn't end in `impl`? `HttpPostingProvisionExecutor` or something maybe?

core/src/test/java/com/novoda/noplayer/exoplayer/ExoPlayerFacadeTest.java
ouchadam: I'm not tooo fussed, it's a test 🍷

PR #56, opened 2017-06-26
ALL-3936/Cues

Mecharyry:
## Problem
We are exposing `ExoPlayer` code in the `NoPlayerView` through the subtitle `Cue`.

## Solution
Create `NoPlayer` specific versions for `Cue` and `SubtitleView` and map between `Cue` and `NoPlayerCue`.

### Test(s) added 
Nope, just updated the already existing tests. 

### Screenshots
No UI changes.

### Paired with 
@ouchadam


core/src/main/java/com/novoda/noplayer/SubtitleView.java
```diff
@@ -0,0 +1,82 @@
...
+import android.graphics.Canvas;
+import android.util.AttributeSet;
+import android.view.View;
+
+import com.novoda.noplayer.model.NoPlayerCue;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public final class SubtitleView extends View {
```
zegnus: Can the client then use any view? If yes, can we create an interface with `setCues`?

core/src/main/java/com/novoda/noplayer/SubtitleView.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        float textSizePx = textSizeType == ABSOLUTE ? textSize
+                : textSize * (textSizeType == FRACTIONAL ? (bottom - top) : (rawBottom - rawTop));
+        if (textSizePx <= 0) {
+            return;
+        }
+
+        for (int i = 0; i < cueCount; i++) {
+            painters.get(i).draw(cues.get(i), applyEmbeddedStyles, applyEmbeddedFontSizes,
```
zegnus: Can you move all the params in different lines? :)

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerCueMapper.java
```diff
@@ -0,0 +1,34 @@
+package com.novoda.noplayer.exoplayer;
+
+import com.google.android.exoplayer2.text.Cue;
+import com.novoda.noplayer.model.NoPlayerCue;
+import com.novoda.noplayer.model.TextCues;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class ExoPlayerCueMapper {
```
zegnus: final and private constructor?

core/src/main/java/com/novoda/noplayer/SubtitleView.java
ouchadam: nope, this view is not changable by a client as it's inflated by `NoPlayerView`

The `PlayerView` interface does have a callback for `setSubtitleCue(TextCues textCues)` which a client can manually implement if they want

zegnus: Still now that you have implemented your own view you can create a contract with
```
void setCues(List<NoPlayerCue> cues);
void setVisibility(int visible);
```

So that we can interchange the view with a different implementation or let the client to use its own view

Mecharyry: Sure we could. Think this is out of the scope of this PR though. Could document with an issue?

zegnus: But this view work wasn't part of this piece of work anyway, adding the interface can be done fast ;)

ouchadam: I'm a little confused, we're not coupled to the `SubtitleView` in any way. In order for a client to provide their own subtitle displaying view they can implement `PlayerView` and that has a callback `setCues(List<NoPlayerCue> cues)`, these are the raw subtitles and any view could be used

core/src/main/java/com/novoda/noplayer/exoplayer/ExoPlayerCueMapper.java
Mecharyry: 9c26b4e

core/src/main/java/com/novoda/noplayer/SubtitleView.java
Mecharyry: Here a3d62e1 and 8d3fdd0

core/src/main/java/com/novoda/noplayer/SubtitleView.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private List<NoPlayerCue> cues;
+
+    public SubtitleView(Context context, AttributeSet attrs) {
+        super(context, attrs);
+        painters = new ArrayList<>();
+    }
+
+    public void setCues(List<NoPlayerCue> cues) {
+        if (this.cues == cues) {
```
juankysoriano: Would be any benefit of doing `equals` instead of `==`? I believe `NoPlayerCue` is implementing `equals()` already, so why not?

core/src/main/java/com/novoda/noplayer/SubtitleView.java
```diff
@@ -0,0 +1,84 @@
...
+        while (painters.size() < cueCount) {
+            painters.add(new SubtitlePainter(getContext()));
+        }
+
+        invalidate();
+    }
+
+    @Override
+    public void dispatchDraw(Canvas canvas) {
+        int cueCount = (cues == null) ? NO_CUES : cues.size();
```
juankysoriano: can we quickly return if `cues == NULL`?

core/src/main/java/com/novoda/noplayer/SubtitleView.java
```diff
@@ -0,0 +1,84 @@
...
+        super(context, attrs);
+        painters = new ArrayList<>();
+    }
+
+    public void setCues(List<NoPlayerCue> cues) {
+        if (this.cues == cues) {
+            return;
+        }
+
+        this.cues = cues;
```
zegnus: Copy the list instead of assigning it, otherwise we loose control on what happens to the element of that list outside of this class

core/src/main/java/com/novoda/noplayer/SubtitleView.java
juankysoriano: Actually, if `cues.isEmpty()` we should also quickly return, right?

core/src/main/java/com/novoda/noplayer/SubtitleView.java
ouchadam: in the case of long piece of content a quick hash check will be a lot faster than chaining `equals()` to all the children

ouchadam: a3ecc5227ec9201bbe9059e473944accd2201cf1 win win!

core/src/main/java/com/novoda/noplayer/SubtitleView.java
ouchadam: a3ecc5227ec9201bbe9059e473944accd2201cf1 we're now using `TextCues` and made the internal list immutable

core/src/main/java/com/novoda/noplayer/SubtitleView.java
ouchadam: 073efabdcf01bbe7e654c80106732d43bc2d9ff2

core/src/main/java/com/novoda/noplayer/SubtitleView.java
zegnus: 👍

core/src/main/java/com/novoda/noplayer/model/TextCues.java
```diff
@@ -39,4 +40,16 @@ public String toString() {
                 "cues=" + cues +
                 '}';
     }
+
+    public int size() {
```
juankysoriano: let's move this before the `equals` and `hashCode`?

Issue #53, opened 2017-06-26
PlayerFactory use a Builder


Mecharyry:
#### Problem
Lot's of "discussions" around the `PlayerFactory` when adding the `DOWNGRADE_SECURE_DECODER` because we passed a boolean all the way down instead of having two different method signatures.

#### Potential Solution
Use the `Builder` pattern, it's ideal when you want to customise something.

#### Impact
Extensible and we can provide easily visible defaults.

Mecharyry:
Pass down `MediaCodecSelector` when doing this.

ouchadam:
fixed by #57

PR #57, opened 2017-06-27
ALL-3936/Player Builder

Mecharyry:
## Problem
As per issue #53 and the conversation that took place in https://github.com/novoda/no-player/pull/52/files/4748c1dd02defdb565cd7174a38deb90ee214d9d#diff-d137d202ec4e30de3f26dfb2c0131896 we want to replace the `PlayerFactory` with a `PlayerBuilder` as the player has a number of customizations that can be exposed to our clients.

## Solution
Create a `PlayerBuilder` and a `PlayerCreator`. We have exposed a `withDowngradedSecureDecoder` to hide the `boolean` we had before. This `boolean` is passed down to the `ExoPlayer` level at which point it is transformed into the necessary `MediaCodecSelector`. The `MediaCodecSelector` is not directly exposed in the `Builder` because it is only necessary for `ExoPlayer`.

### Test(s) added 
Updating tests that we broke as a result of these changes.

### Screenshots
No UI changes.

### Paired with 
@ouchadam

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/exoplayer/NoPlayerExoPlayerCreator.java
+++ b/core/src/main/java/com/novoda/noplayer/exoplayer/NoPlayerExoPlayerCreator.java
@@ -3,6 +3,7 @@ package com.novoda.noplayer.exoplayer;
 import android.content.Context;
 import android.os.Handler;
 
+import com.google.android.exoplayer2.mediacodec.MediaCodecSelector;
 import com.google.android.exoplayer2.trackselection.DefaultTrackSelector;
 import com.google.android.exoplayer2.trackselection.FixedTrackSelection;
 import com.google.android.exoplayer2.upstream.DefaultDataSourceFactory;
@@ -51,6 +52,8 @@ public class NoPlayerExoPlayerCreator {
 
             DefaultTrackSelector trackSelector = new DefaultTrackSelector();
 
+            MediaCodecSelector mediaCodecSelector = downgradeSecureDecoder ? SecurityDowngradingCodecSelector.newInstance() : MediaCodecSelector.DEFAULT;
+
             ExoPlayerTrackSelector exoPlayerTrackSelector = ExoPlayerTrackSelector.newInstance(trackSelector);
             FixedTrackSelection.Factory trackSelectionFactory = new FixedTrackSelection.Factory();
             ExoPlayerAudioTrackSelector exoPlayerAudioTrackSelector = new ExoPlayerAudioTrackSelector(exoPlayerTrackSelector, trackSelectionFactory);
@@ -81,7 +84,7 @@ public class NoPlayerExoPlayerCreator {
                     loadTimeout,
                     heart,
                     drmSessionCreator,
-                    downgradeSecureDecoder
+                    mediaCodecSelector
             );
         }
     }
--- a/core/src/main/java/com/novoda/noplayer/exoplayer/SecurityDowngradingCodecSelector.java
+++ b/core/src/main/java/com/novoda/noplayer/exoplayer/SecurityDowngradingCodecSelector.java
@@ -6,6 +6,8 @@ import com.google.android.exoplayer2.mediacodec.MediaCodecUtil;
 
 class SecurityDowngradingCodecSelector implements MediaCodecSelector {
 
+    private static final boolean USE_INSECURE_DECODER = false;
+
     private final InternalMediaCodecUtil internalMediaCodecUtil;
 
     public static SecurityDowngradingCodecSelector newInstance() {
@@ -20,7 +22,7 @@ class SecurityDowngradingCodecSelector implements MediaCodecSelector {
     @Override
     public MediaCodecInfo getDecoderInfo(String mimeType, boolean contentRequiresSecureDecoder)
             throws MediaCodecUtil.DecoderQueryException {
-        return internalMediaCodecUtil.getDecoderInfo(mimeType, false);
+        return internalMediaCodecUtil.getDecoderInfo(mimeType, USE_INSECURE_DECODER);
     }
 
     @Override
--- a/core/src/main/java/com/novoda/noplayer/exoplayer/SecurityDowngradingCodecSelector.java
+++ b/core/src/main/java/com/novoda/noplayer/exoplayer/SecurityDowngradingCodecSelector.java
@@ -7,22 +7,20 @@ import com.google.android.exoplayer2.mediacodec.MediaCodecUtil;
 class SecurityDowngradingCodecSelector implements MediaCodecSelector {
 
     private final InternalMediaCodecUtil internalMediaCodecUtil;
-    private final boolean downgradeSecureDecoder;
 
-    public static SecurityDowngradingCodecSelector newInstance(boolean downgradeSecureDecoder) {
+    public static SecurityDowngradingCodecSelector newInstance() {
         InternalMediaCodecUtil internalMediaCodecUtil = new InternalMediaCodecUtil();
-        return new SecurityDowngradingCodecSelector(internalMediaCodecUtil, downgradeSecureDecoder);
+        return new SecurityDowngradingCodecSelector(internalMediaCodecUtil);
     }
 
-    SecurityDowngradingCodecSelector(InternalMediaCodecUtil internalMediaCodecUtil, boolean downgradeSecureDecoder) {
+    SecurityDowngradingCodecSelector(InternalMediaCodecUtil internalMediaCodecUtil) {
         this.internalMediaCodecUtil = internalMediaCodecUtil;
-        this.downgradeSecureDecoder = downgradeSecureDecoder;
     }
 
     @Override
     public MediaCodecInfo getDecoderInfo(String mimeType, boolean contentRequiresSecureDecoder)
             throws MediaCodecUtil.DecoderQueryException {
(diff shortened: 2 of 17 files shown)


core/src/main/java/com/novoda/noplayer/exoplayer/NoPlayerExoPlayerCreator.java
```diff
@@ -51,6 +52,8 @@ ExoPlayerTwoImpl create(Context context, DrmSessionCreator drmSessionCreator, bo
 
             DefaultTrackSelector trackSelector = new DefaultTrackSelector();
 
+            MediaCodecSelector mediaCodecSelector = downgradeSecureDecoder ? SecurityDowngradingCodecSelector.newInstance() : MediaCodecSelector.DEFAULT;
```
Mecharyry: `boolean downgradeSecureDecoder` is transformed into the associated `MediaCodecSelector`.

core/src/main/java/com/novoda/noplayer/player/NoPlayerCreator.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.novoda.noplayer.Player;
+import com.novoda.noplayer.drm.DrmHandler;
+import com.novoda.noplayer.drm.DrmType;
+import com.novoda.noplayer.exoplayer.NoPlayerExoPlayerCreator;
+import com.novoda.noplayer.exoplayer.drm.DrmSessionCreator;
+import com.novoda.noplayer.exoplayer.drm.DrmSessionCreatorFactory;
+import com.novoda.noplayer.mediaplayer.NoPlayerMediaPlayerCreator;
+
+class NoPlayerCreator {
```
Mecharyry: This is the old `PlayerFactory` minus a few wrapper methods, we only expose the one method create and it is used only in the `PlayerBuilder`.

core/src/main/java/com/novoda/noplayer/player/PlayerBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+import com.novoda.noplayer.exoplayer.NoPlayerExoPlayerCreator;
+import com.novoda.noplayer.exoplayer.drm.DrmSessionCreatorFactory;
+import com.novoda.noplayer.mediaplayer.NoPlayerMediaPlayerCreator;
+import com.novoda.utils.AndroidDeviceVersion;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class PlayerBuilder {
```
Mecharyry: `Builder` party 🎉

core/src/main/java/com/novoda/noplayer/player/PlayerBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+        prioritizedPlayerTypes = new PrioritizedPlayerTypes(types);
+        return this;
+    }
+
+    public PlayerBuilder withDowngradedSecureDecoder() {
+        downgradeSecureDecoder = true;
+        return this;
+    }
+
+    public Player build(Context context) {
```
juankysoriano: as a user of the library I can still use `NoPlayerExoPlayerCreator` directly and create it without the builder. 

Is this intentional?

ouchadam: I don't see how to avoid this without flattening all the packages 😢

core/src/main/java/com/novoda/noplayer/exoplayer/SecurityDowngradingCodecSelector.java
```diff
@@ -7,22 +7,20 @@
...
     }
 
-    SecurityDowngradingCodecSelector(InternalMediaCodecUtil internalMediaCodecUtil, boolean downgradeSecureDecoder) {
+    SecurityDowngradingCodecSelector(InternalMediaCodecUtil internalMediaCodecUtil) {
         this.internalMediaCodecUtil = internalMediaCodecUtil;
-        this.downgradeSecureDecoder = downgradeSecureDecoder;
     }
 
     @Override
     public MediaCodecInfo getDecoderInfo(String mimeType, boolean contentRequiresSecureDecoder)
```
juankysoriano: is `contentRequiresSecureDecoder` not used anymore? if so... kill it? (if swe can, which I don't know)

core/src/main/java/com/novoda/noplayer/player/PlayerBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public PlayerBuilder withPriority(PlayerType playerType, PlayerType... playerTypes) {
+        List<PlayerType> types = new ArrayList<>();
+        types.add(playerType);
+        types.addAll(Arrays.asList(playerTypes));
+        prioritizedPlayerTypes = new PrioritizedPlayerTypes(types);
+        return this;
+    }
+
+    public PlayerBuilder withDowngradedSecureDecoder() {
```
juankysoriano: 👍

core/src/main/java/com/novoda/noplayer/exoplayer/SecurityDowngradingCodecSelector.java
ouchadam: it's from `ExoPlayer MediaCodecSelector`, we can't remove it, unless you mean rename it?

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -50,10 +48,12 @@ protected void onCreate(Bundle savedInstanceState) {
     protected void onStart() {
         super.onStart();
         // TODO: Add switch in UI to avoid redeploy.
-        PlayerFactory playerFactory = PlayerFactory.newInstance(this, PrioritizedPlayerTypes.prioritizeExoPlayer());
         DataPostingModularDrm drmHandler = new DataPostingModularDrm(EXAMPLE_MODULAR_LICENSE_SERVER_PROXY);
 
-        player = playerFactory.create(DrmType.WIDEVINE_MODULAR_STREAM, drmHandler);
+        player = new PlayerBuilder()
```
juankysoriano: To be honest I feel the builder a super cool idea! 💯

ouchadam: you can thank @Mecharyry trolling me and @zegnus 😂

core/src/main/java/com/novoda/noplayer/exoplayer/NoPlayerExoPlayerCreator.java
ouchadam: @joetimmins the boolean now lives on `PlayerBuilder` -> `NoPlayerCreator` and finally here where it gets converted to the codec selector

core/src/main/java/com/novoda/noplayer/exoplayer/SecurityDowngradingCodecSelector.java
juankysoriano: I didn't notice it was from the exoplayer one, ok then 👍

core/src/main/java/com/novoda/noplayer/player/PlayerBuilder.java
juankysoriano: 😢 

I have taken a look and I don't see either

core/src/main/java/com/novoda/noplayer/exoplayer/SecurityDowngradingCodecSelector.java
```diff
@@ -7,22 +7,20 @@
...
             throws MediaCodecUtil.DecoderQueryException {
-        return internalMediaCodecUtil.getDecoderInfo(mimeType, contentRequiresSecureDecoder && !downgradeSecureDecoder);
+        return internalMediaCodecUtil.getDecoderInfo(mimeType, false);
```
zegnus: can you extract that boolean? I have no idea what's false there :P

core/src/main/java/com/novoda/noplayer/player/PlayerBuilder.java
zegnus: Can `NoPlayerCreator` be properly used by the client? if it doesn't make sense for the client to do so, we might have to flatten the packages... java, libraries and packages are not friends, it really sucks :S

zegnus:
💯  niiice!

core/src/main/java/com/novoda/noplayer/player/PlayerBuilder.java
ouchadam: `NoPlayerCreator` is not public, it's the individual player creators which are, like `NoPlayerExoPlayerCreator`

I don't think it's problem or enough of a reason to flatten the entire library. 

There's no issue or side effects of a client creating a single player directly via the public player creators, if anything it's an architectural feature. The `PlayerBuilder` is syntactic sugar around selecting a player for given types.

Mecharyry: It's the `NoPlayerMediaPlayerCreator` and `NoPlayerExoPlayerCreator` that are exposed. They are the sole entry points into the `exo-player` and `media-player` packages. Flattening would solve this problem but make maintaining `no-player` extremely painful.

zegnus: Agreed offline, we'll expose these creators as part of the public API

PR #58, opened 2017-06-27
ALL-3936/JavaDoc

Mecharyry:
## Problem
As per issue #40 JavaDoc public APIs. We have very little information on any of our public APIs to aid clients of this library.

## Solution
Add JavaDoc to `Player` and `PlayerBuilder` the two public APIs.

### Test(s) added 
No, adding JavaDoc only.

### Screenshots
No UI changes.

### Paired with 
@ouchadam


core/src/main/java/com/novoda/noplayer/Player.java
```diff
@@ -14,38 +14,121 @@
 
 public interface Player extends PlayerState {
 
+    /**
```
juankysoriano: I think that 
```
/**
* Plays content of a prepared {@link Player}.
* @see Player.PreparedListener
*/
``` 
Would read much better. also for the others

core/src/main/java/com/novoda/noplayer/Player.java
```diff
@@ -14,38 +14,121 @@
...
     void stop();
 
+    /**
+     * Stops playback of content and drops all internal resources. The instance of {@link Player} should not be
+     * used after calling {@link Player#release()}.
+     */
     void release();
 
+    /**
+     * Loads the video content and triggers the {@link Player.PreparedListener}.
```
juankysoriano:
```
/**
* Loads video content and prepares the player. 
* @see Player.PreparedListener 
*/
```

Same for the other

core/src/main/java/com/novoda/noplayer/player/PlayerBuilder.java
```diff
@@ -52,12 +91,26 @@ public PlayerBuilder withPriority(PlayerType playerType, PlayerType... playerTyp
...
         return this;
     }
 
-    public Player build(Context context) {
+    /**
+     * Builds a new Player instance.
+     *
+     * @param context
+     * @return a Player instance.
+     * @throws UnableToCreatePlayerException thrown when the configuration is not supported and there is no way to recover.
```
zegnus: :)

Mecharyry:
I guess quite a few of the `Player` methods can throw, if the client doesn't make sure the `Player` is prepared before calling some methods.

zegnus:
@Mecharyry do you think the client might benefit on knowing that not only the method won't work properly, but that an exception will be thrown

Mecharyry:
I'm not sure. We highlight that the `Player` has to be prepared before calling the method. I think adding the `exception` would just be repeating ourselves. I'm not sure though.

zegnus:
Do we have any disadvantage documenting the exception?

Mecharyry:
No, we should document it I think. Better to be overly verbose on a JavaDoc than miss something out and have issues on the library.

zegnus:
👍

Mecharyry:
@zegnus I've tried adding the throws for all of the methods on the `Player` but I'm not sure this is doable. Here's my reasoning.

- We are throwing an `UncheckedException` which we aren't expecting clients to catch and handle. It represents a failure of state based on how they are using the library. 
- The underlying libraries although based on the same interface could be initialised in very different ways and what may cause a crash in one does not necessarily mean it will crash in the other.

zegnus:
@Ryan I'll reply in order:

- The doc is not for expecting a client to handle it, but to be aware what happens with that method. Although the java conventions doesn't mention that you have to document all the unchecked exceptions, they are properly documented in the majority of the Java Standard Libraries (example: https://docs.oracle.com/javase/7/docs/api/java/util/List.html#get(int))

- As a user, if I code against an interface I would expect to have to call the methods in the same order regardless of the internal implementation. Even if its not the case, makes the public API much simpler to just provide one way to perform an action. You can also elaborate in which case an exception is thrown in the doc

core/src/main/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -261,33 +260,28 @@ public void detach(PlayerView playerView) {
...
     @Override
-    public boolean selectAudioTrack(PlayerAudioTrack audioTrack) {
+    public boolean selectAudioTrack(PlayerAudioTrack audioTrack) throws IllegalStateException {
         return mediaPlayer.selectAudioTrack(audioTrack);
     }
 
     @Override
     public boolean showSubtitleTrack(PlayerSubtitleTrack subtitleTrack) {
-        SubtitlesError subtitlesError = new SubtitlesError("Subtitles not implemented for Android Media Player", new IllegalStateException());
-        listenersHolder.getErrorListeners().onError(this, subtitlesError);
```
zegnus: Are you sure we want to remove the error report in these methods?

Mecharyry: Yes. After talking to Adam, it's not worth reporting in this way. It's not supported in `MediaPlayer`but it should be in the future. I would like to have a wider array of `Capabilities` that `Player`s can support and then we select a `Player` based on that.

zegnus:
We'll move some classes to an `internal` package and update documentation on those so its clear from the client

core/src/test/java/com/novoda/noplayer/mediaplayer/AndroidMediaPlayerFacadeTest.java
```diff
@@ -335,6 +336,18 @@ public void givenMediaPlayerIsNotPlaying_whenPausing_thenDoesNotPausesMediaPlaye
     }
 
     @Test
+    public void givenMediaPlayerIsNotInPlaybackState_whenPausing_thenThrowsIllegalStateException() {
```
juankysoriano: 👍

PR #59, opened 2017-06-28
ALL-3936/Internal package

Mecharyry:
## Problem
As per the discussion in #58 we want to expose an `internal` package for classes that we are forced to make public but should not be used by clients and therefore do not require JavaDocs.

## Solution
Create an `internal` package and move internally public classes there!

Replace the getters for `ErrorListeners` etc in the `ListenersHolder` so that they use the interface rather than the implementation. Allows us to make those implementations package protected so that the only way to add and remove is through the `ListenersHolder`.

Remove generics on the `Heartbeat.Callbacks` we always need a `Player` anyway.

Add the `Heartbeat.Callback` to the `Player` because that's where all the other `Callbacks` and `Listeners` are always declared here.

### Test(s) added 
No, just updating the broken packages.

### Screenshots
No UI changes.

### Paired with 
@ouchadam


core/src/main/java/com/novoda/noplayer/Player.java
```diff
@@ -7,13 +7,14 @@
...
 import com.novoda.noplayer.model.VideoPosition;
-import com.novoda.noplayer.player.PlayerInformation;
 
 import java.util.List;
 import java.util.Map;
 
 public interface Player extends PlayerState {
 
+
+
```
juankysoriano: oooops

ouchadam: 👍

core/src/main/java/com/novoda/noplayer/Listeners.java
```diff
@@ -30,9 +30,9 @@
 
     void removeBitrateChangedListener(Player.BitrateChangedListener bitrateChangedListener);
 
-    void addHeartbeatCallback(Heart.Heartbeat.Callback<Player> callback);
+    void addHeartbeatCallback(Player.HeartbeatCallback heartbeatCallback);
```
Mecharyry: Heart.Heartbeat.Callback has been moved to the Player to match the other listeners that the Player supports.

core/src/main/java/com/novoda/noplayer/NoPlayerCreator.java
```diff
@@ -44,8 +46,12 @@ private Player createPlayerForType(PlayerType playerType, DrmType drmType, DrmHa
             case MEDIA_PLAYER:
                 return noPlayerMediaPlayerCreator.createMediaPlayer(context);
             case EXO_PLAYER:
-                DrmSessionCreator drmSessionCreator = drmSessionCreatorFactory.createFor(drmType, drmHandler);
-                return noPlayerExoPlayerCreator.createExoPlayer(context, drmSessionCreator, downgradeSecureDecoder);
+                try {
+                    DrmSessionCreator drmSessionCreator = drmSessionCreatorFactory.createFor(drmType, drmHandler);
+                    return noPlayerExoPlayerCreator.createExoPlayer(context, drmSessionCreator, downgradeSecureDecoder);
+                } catch (DrmSessionCreatorException exception) {
+                    throw new UnableToCreatePlayerException(exception);
```
Mecharyry: Mapping an `internal` exception to a public facing exception.

core/src/main/java/com/novoda/noplayer/Player.java
```diff
@@ -225,4 +224,9 @@ public void onLoadTimeout() {
 
         void onLoadTimeout();
     }
+
+    interface HeartbeatCallback {
```
Mecharyry: Moved from `Heart.Heartbeat.Callback` to match the other callbacks on the `Player`.

core/src/main/java/com/novoda/noplayer/drm/StreamingModularDrm.java
```diff
@@ -3,4 +3,23 @@
 public interface StreamingModularDrm extends DrmHandler {
 
     byte[] executeKeyRequest(ModularDrmKeyRequest request) throws DrmRequestException;
+
+    final class DrmRequestException extends Exception {
```
Mecharyry: Only used for `StreamingModularDrm` so inlined on the interface.

core/src/main/java/com/novoda/noplayer/internal/Heart.java
```diff
@@ -63,24 +65,19 @@ public boolean isBeating() {
         return beating;
     }
 
-    public static class Heartbeat<T> implements Runnable {
```
Mecharyry: No need for the generic, we always pass back a `Player`.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/EventInfoForwarder.java
```diff
@@ -1,21 +1,21 @@
...
 import com.google.android.exoplayer2.trackselection.TrackSelectionArray;
-import com.novoda.noplayer.listeners.InfoListeners;
+import com.novoda.noplayer.Player;
 
 import java.util.HashMap;
 
 class EventInfoForwarder implements ExoPlayer.EventListener {
 
-    private final InfoListeners infoListeners;
+    private final Player.InfoListener infoListener;
```
Mecharyry: Coding to the exposed interface rather than the internal class.

core/src/main/java/com/novoda/noplayer/internal/listeners/PlayerListenersHolder.java
```diff
@@ -125,39 +119,39 @@ public void removeVideoSizeChangedListener(Player.VideoSizeChangedListener video
         videoSizeChangedListeners.remove(videoSizeChangedListener);
     }
 
-    public ErrorListeners getErrorListeners() {
+    public Player.ErrorListener getErrorListeners() {
```
Mecharyry: Use the interfaces, so that we can make the `ErrorListeners` etc package protected.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/TrackType.java
```diff
@@ -0,0 +1,7 @@
+package com.novoda.noplayer.internal.exoplayer.mediasource;
+
+enum TrackType {
```
jackSzm: Where is this coming from ?

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -180,7 +180,7 @@ public void seekTo(VideoPosition position) throws IllegalStateException {
     }
 
     @Override
-    public void pause() {  // TODO: throw to match ExoPlayer?
+    public void pause() throws IllegalStateException {
```
jackSzm: 💯

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/forwarder/BufferHeartbeatListener.java
```diff
@@ -0,0 +1,23 @@
+package com.novoda.noplayer.internal.mediaplayer.forwarder;
+
+import com.novoda.noplayer.Player;
+import com.novoda.noplayer.internal.mediaplayer.CheckBufferHeartbeatCallback;
+
+class BufferHeartbeatListener implements CheckBufferHeartbeatCallback.BufferListener {
```
jackSzm: Is this moved from somewhere else as well ?

Mecharyry: Yes, this was in the `mediaplayer.forwarder` package before. Now it's been moved to `internal.mediaplayer.forwarder`.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/TrackType.java
Mecharyry: Just moved from `exoplayer.mediasource` to `internal.exoplayer.mediasource`.

PR #61, opened 2017-06-29
Subtitles crash and demo flow

Mecharyry:
## Problem
Crash when trying to check equality of equals on a object that was null 😬 

`NoPlayer` Demo doesn't exactly follow the flow we expect clients to implement...

## Solution
Swap the equality check to prevent crashes.

Make `NoPlayer` demo follow the flow we expect when creating and interacting with a `Player`.

### Test(s) added 
No, just changing flow in Activities.

### Screenshots
No UI changes.

### Paired with 
@Dorvaryn

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/SubtitleView.java
+++ b/core/src/main/java/com/novoda/noplayer/SubtitleView.java
@@ -19,7 +19,6 @@ public final class SubtitleView extends View {
     private static final boolean APPLY_EMBEDDED_FONT_STYLES = true;
 
     private static final int ZERO_PIXELS = 0;
-    private static final int NO_CUES = 0;
 
     private final List<SubtitlePainter> painters;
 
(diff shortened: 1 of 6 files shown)


ouchadam:
💯

core/src/main/java/com/novoda/noplayer/SubtitleView.java
```diff
@@ -31,12 +30,12 @@ public SubtitleView(Context context, AttributeSet attrs) {
     }
 
     public void setCues(TextCues textCues) {
```
joetimmins: Can this `textCues` ever be null? Because if it can, line 37 can throw `NullPointerException`.

Mecharyry: 💯  We've updated the mapper, that passes to this, to ensure that it always returns a list.

PR #62, opened 2017-06-29
Hiding `PlayerChecker`

ouchadam:
## Problem

We're currently exposing the `PlayerChecker`and `AndroidMediaPlayerType` which attempts to read the systems internal media player implementation.

## Solution

The internal media player name is now exposed via the `PlayerInformation.getName`. If the internal type is required we can add multiple player information impls for each internal type.

### Test(s) added 

Yep around the `PlayerInformation` impls

### Screenshots

no ui changes

### Paired with 

noone

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/InternalMediaPlayerTypeReader.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/InternalMediaPlayerTypeReader.java
@@ -2,7 +2,7 @@ package com.novoda.noplayer.internal.mediaplayer;
 
 import android.os.Build;
 
-class PlayerChecker {
+class InternalMediaPlayerTypeReader {
 
     private static final String PROP_USE_NU_PLAYER = "media.stagefright.use-nuplayer";
     private static final String PROP_USE_AWESOME_PLAYER_PERSIST = "persist.sys.media.use-awesome";
@@ -11,7 +11,7 @@ class PlayerChecker {
     private final int deviceOSVersion;
     private final SystemProperties systemProperties;
 
-    PlayerChecker(SystemProperties systemProperties, int deviceOSVersion) {
+    InternalMediaPlayerTypeReader(SystemProperties systemProperties, int deviceOSVersion) {
         this.systemProperties = systemProperties;
         this.deviceOSVersion = deviceOSVersion;
     }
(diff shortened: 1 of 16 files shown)


core/src/main/java/com/novoda/noplayer/internal/mediaplayer/InternalMediaPlayerTypeReader.java
```diff
@@ -2,7 +2,7 @@
 
 import android.os.Build;
 
-public class PlayerChecker {
+class InternalMediaPlayerTypeReader {
```
zegnus: Now that's not public can we just call it `MediaPlayerTypeReader` ?

zegnus:
Small comment, LGTM 👍

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/InternalMediaPlayerTypeReader.java
ouchadam: I wasn't sure, because this is the internal media players internal player (nu player/awesome player) 🎃

happy to change

core/src/test/java/com/novoda/noplayer/internal/mediaplayer/MediaPlayerInformationTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @Test
+    public void whenReadingVersion_thenReturnsExoPlayerLibraryVersion() {
+
+        String version = playerInformation.getVersion();
+
+        assertThat(version).isEqualTo(Build.VERSION.RELEASE);
+    }
+
+    @Test
+    public void whenPlayerType_thenReturnsExoPlayer() {
```
juankysoriano: wrong test name?

core/src/test/java/com/novoda/noplayer/internal/mediaplayer/MediaPlayerInformationTest.java
```diff
@@ -0,0 +1,56 @@
...
+    public void givenInternalNuPlayer_whenReadingName_thenReturnsMediaPlayerNuPlayer() {
+        given(playerTypeReader.getPlayerType()).willReturn(AndroidMediaPlayerType.NU);
+
+        String name = playerInformation.getName();
+
+        assertThat(name).isEqualTo("MediaPlayer: NuPlayer");
+    }
+
+    @Test
+    public void whenReadingVersion_thenReturnsExoPlayerLibraryVersion() {
```
juankysoriano: Is the exo player here right?

core/src/test/java/com/novoda/noplayer/internal/mediaplayer/MediaPlayerInformationTest.java
ouchadam: oops!

bad copy paste :

PR #64, opened 2017-06-30
ALL-3936/Error classes

Mecharyry:
## Problem
We are exposing the internals of `no-player` through our `Player.PlayerError` implementations that reference directly media-player or exo-player.

## Solution
Remove the specific implementations in favour of an `PlayerErrorType` that lives on the `Player.PlayerError`. We need to make these types more generic to apply across `Player` boundaries. This is the first step to remove the class implementations for the type. 

### Test(s) added 
No, just swapping out the implementations for a generic exception.

### Screenshots
No UI changes.

### Paired with 
@Dorvaryn


core/src/main/java/com/novoda/noplayer/NoPlayerError.java
```diff
@@ -0,0 +1,22 @@
+package com.novoda.noplayer;
+
+public class NoPlayerError implements Player.PlayerError {
+
+    private final PlayerErrorType playerErrorType;
```
zegnus: 💯

core/src/main/java/com/novoda/noplayer/PlayerErrorType.java
```diff
@@ -0,0 +1,17 @@
...
+
+public enum PlayerErrorType {
+
+    INVALID_RESPONSE_CODE,
+    MALFORMED_CONTENT,
+    FAILED_DRM_DECRYPTION,
+    FAILED_DRM_INITIATING,
+    FAILED_DRM_REQUEST,
+    CONNECTIVITY_ERROR,
+    MEDIA_FORMAT_NOT_RECOGNISED,
```
ataulm: RECOGNIZED :us:

core/src/main/java/com/novoda/noplayer/PlayerErrorType.java
```diff
@@ -0,0 +1,17 @@
...
+    MEDIA_SERVER_DIED,
+    MEDIA_SOURCE_ERROR,
+    STREAMED_VIDEO_ERROR,
+    UNKNOWN
```
ataulm: @Dorvaryn disappointed to not find `UNKOWN`

PR #65, opened 2017-06-30
Delegate SurfaceHolder lifecycle to ExoPlayer  

ouchadam:
## Problem

Spamming play/pause causes the player to freeze and eventually get stuck. This is due to us clearing and resetting the player `SurfaceHolder` each time `Player.play` is called 

## Solution

`SimpleExoPlayer` internally handles the lifecycle of the `SurfaceHolder` so we can simply attach to `SurfaceHolder` to it as part of `Player.attach`.

- Also added a `remove` to the `SurfaceHolderRequest` to avoid leaking any callbacks

### Test(s) added 

Yep around the attach of the `SurfaceHolder` to the `ExoPlayer` via the `Facade` and `Impl`

### Screenshots

no changes in the demo

### Paired with 

nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
@@ -22,12 +22,15 @@ import com.novoda.noplayer.model.Timeout;
 import com.novoda.noplayer.model.VideoDuration;
 import com.novoda.noplayer.model.VideoPosition;
 
+import java.util.ArrayList;
 import java.util.List;
 
 class AndroidMediaPlayerImpl implements Player {
 
     private static final VideoPosition NO_SEEK_TO_POSITION = VideoPosition.INVALID;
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -255,12 +259,20 @@ public void attach(PlayerView playerView) {
 
     @Override
     public void detach(PlayerView playerView) {
+        clearSurfaceHolderCallbacks();
         surfaceHolderRequester = null;
```
Mecharyry: Move this into the `clearSurfaceHolderCallbacks`?

ouchadam: I wasn't sure about it, as it meant the method was doing 2 things and kinda hid the nulling. 

I would expect a detach method to null stuff up. I'm easy though

ataulm:
LGTM :99:

PR #66, opened 2017-06-30
Decouple error messaging

Mecharyry:
## Problem
An error message needs to be exposed in `NoPlayerError` but its representation changes based on the underlying player. For example:

MediaPlayer -> onError(int what, int extra) -> which when transformed into `new Throwable(extra)` left out the `what` which could be important to client applications.

ExoPlayer -> just passed back the original exception as the cause.

## Solution
Drop `Throwable getCause()` and replace it with `String getMessage()` where we create an `ErrorFormatting` which defines the format of the error message based on the underlying player. 

### Test(s) added 
Yes, added tests for the `ErrorFormatting` classes.

### Screenshots
No UI changes.

### Paired with 
@Dorvaryn

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ErrorFormating.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ErrorFormating.java
@@ -1,0 +1,12 @@
+package com.novoda.noplayer.internal.exoplayer.forwarder;
+
+final class ErrorFormating {
+
+    private ErrorFormating() {
+    }
+
+    static String formatMessage(Throwable throwable) {
+        return throwable.getClass().getName() + ": " + throwable.getMessage();
+    }
+
+}
--- a/core/src/main/java/com/novoda/noplayer/NoPlayerError.java
+++ b/core/src/main/java/com/novoda/noplayer/NoPlayerError.java
@@ -3,20 +3,20 @@ package com.novoda.noplayer;
 public class NoPlayerError implements Player.PlayerError {
 
     private final PlayerErrorType playerErrorType;
-    private final Throwable cause;
+    private final String message;
 
-    public NoPlayerError(PlayerErrorType playerErrorType, Throwable cause) {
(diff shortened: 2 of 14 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ErrorFormating.java
```diff
@@ -0,0 +1,12 @@
+package com.novoda.noplayer.internal.exoplayer.forwarder;
+
+final class ErrorFormating {
+
+    private ErrorFormating() {
+    }
+
+    static String formatMessage(Throwable throwable) {
```
ouchadam: 👍

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ErrorFormating.java
```diff
@@ -0,0 +1,12 @@
+package com.novoda.noplayer.internal.exoplayer.forwarder;
+
+final class ErrorFormating {
```
ouchadam: `ErrorFormating` or `ErrorFormatter`?

Mecharyry: I prefer `ErrorFormatter` 😄

Dorvaryn: Argh 2 against 1 lol.  I don't like `Formatter` for classes that are not instances, as the class itself is not a formatter but a holder for formatting methods

ouchadam: I agree but this class feels like it's in the middle 😛, not quite a holder and not quite a formatter

if we had `interface Formating -> format()` and `ErrorFormating implements Formating` then I'd be 100% onboard, I guess that's pretty overkill for what we need 👻

PR #70, opened 2017-07-13
Only Play when transitioning 

Mecharyry:
## Problem
We had a crash in a client app that occurred because we were double calling `onVideoPlaying` this occurs only on `MediaPlayer` which calls `play` when performing a seek.

## Solution
Store the `isPlaying` state before calling `mediaPlayer.start` and if the state is already `PLAYING` then do not call `onVideoPlaying`.

### Test(s) added 
Added a test to ensure that we do not call `onVideoPlaying` when the `Player` is already playing.

### Screenshots
No UI changes.

### Paired with 
Nobody but @ouchadam found the problem 😂

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
@@ -121,8 +121,11 @@ class AndroidMediaPlayerImpl implements Player {
         requestSurface(new SurfaceHolderRequester.Callback() {
             @Override
             public void onSurfaceHolderReady(SurfaceHolder surfaceHolder) {
+                boolean videoWasNotPlaying = !isPlaying();
                 mediaPlayer.start(surfaceHolder);
-                listenersHolder.getStateChangedListeners().onVideoPlaying();
+                if (videoWasNotPlaying) {
+                    listenersHolder.getStateChangedListeners().onVideoPlaying();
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -121,8 +121,11 @@ public void play() throws IllegalStateException {
         requestSurface(new SurfaceHolderRequester.Callback() {
             @Override
             public void onSurfaceHolderReady(SurfaceHolder surfaceHolder) {
+                boolean videoWasNotPlaying = !isPlaying();
```
juankysoriano: can we move it just before the usage?

juankysoriano:
@Mecharyry super small comment and 🚢  after that

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
Mecharyry: no, we can't. The state gets changed in the next line.

We are storing the `isPlaying` before starting the player internally, which will change the state of `isPlaying` and will result in emitting the `onVideoPlaying`.

juankysoriano: aaah

PR #71, opened 2017-07-13
Reset state on error

Mecharyry:
## Problem
When receiving an error we never change state. If retrying, it is possible to become stuck in a loop because we are saying that the player is still able to play when it is not.

## Solution
Reset `Player` state when an error occurs. 

### Test(s) added 
Yes, verifying that we reset the state in the event of an error.

### Screenshots
No UI changes.

### Paired with 
@Dorvaryn and @ouchadam

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java
@@ -77,7 +77,7 @@ class ExoPlayerTwoImpl implements Player {
         listenersHolder.addErrorListener(new ErrorListener() {
             @Override
             public void onError(PlayerError error) {
-                loadTimeout.cancel();
+                reset();
             }
         });
         listenersHolder.addVideoSizeChangedListener(new VideoSizeChangedListener() {
@@ -150,17 +150,17 @@ class ExoPlayerTwoImpl implements Player {
     @Override
     public void stop() {
         reset();
+        listenersHolder.getStateChangedListeners().onVideoStopped();
     }
 
     @Override
     public void release() {
-        reset();
+        stop();
         listenersHolder.clear();
     }
 
     private void reset() {
         listenersHolder.resetPreparedState();
-        listenersHolder.getStateChangedListeners().onVideoStopped();
         loadTimeout.cancel();
         heart.stopBeatingHeart();
         exoPlayer.release();
@@ -169,7 +169,7 @@ class ExoPlayerTwoImpl implements Player {
     @Override
     public void loadVideo(final Uri uri, final ContentType contentType) {
         if (exoPlayer.hasPlayedContent()) {
-            reset();
+            stop();
         }
         surfaceHolderRequester.removeCallback(onSurfaceReadyCallback);
         onSurfaceReadyCallback = new SurfaceHolderRequester.Callback() {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -150,17 +150,17 @@ public void seekTo(VideoPosition position) throws IllegalStateException {
     @Override
     public void stop() {
         reset();
+        listenersHolder.getStateChangedListeners().onVideoStopped();
```
jackSzm: I guess this is sth else , that is not mentioned in the description of the PR 🤔  ?

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -54,7 +54,7 @@
                            MediaPlayerForwarder forwarder,
                            PlayerListenersHolder listenersHolder,
                            CheckBufferHeartbeatCallback bufferHeartbeatCallback,
-                           LoadTimeout loadTimeoutParam,
+                           LoadTimeout loadTimeout,
```
jackSzm: 💯

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -150,17 +150,17 @@ public void seekTo(VideoPosition position) throws IllegalStateException {
...
     }
 
     @Override
     public void release() {
-        reset();
+        stop();
```
juankysoriano: are we positive we'll always want to do `stop()` when releasing, even if in the future our stop routine changes? or is it just mere coincidence that right now we do the same stuff? I would be cautious when calling a public method inside another. If you are already considering this then 👍

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java
Mecharyry: Sorry, this was in the reset before but actually we want the `reset()` in more places without calling the `onVideoStopped`.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java
Mecharyry: We just had a conversation about it 😆 we should be ok, we have tests backing everything in this class and if we ever wanted to change the implementation of `stop()` we can also change `release()` at the same time. Not that we think we will anyways.

juankysoriano: I was sure you considered it!! 💯  good stuff!

core/src/test/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImplTest.java
```diff
@@ -114,7 +114,7 @@ public void givenPlayerIsInitialised_whenVideoIsPrepared_thenCancelsTimeout() {
         }
 
         @Test
-        public void givenPlayerIsInitialised_whenVideoHasError_thenCancelsTimeout() {
+        public void givenPlayerIsInitialised_whenVideoHasError_thenPlayerResourcesAreReleased_andNotListeners() {
```
juankysoriano: Given the change we have done it seems like we should check that `listenersHolder.getStateChangedListeners().onVideoStopped()` is not called on error? 

Same for the other test

Mecharyry: Thanks for mentioning this @juankysoriano, done now 9b5a02d

PR #73, opened 2017-07-24
Update to latest ExoPlayer

Mecharyry:
## Problem
We are a few patches out of sync with `ExoPlayer` see [here](https://github.com/google/ExoPlayer/blob/dev-v2/RELEASENOTES.md). This should help us resolve an issue we are seeing when switching renditions on a client application.

## Solution
Update `ExoPlayer` dependency to the latest.

### Test(s) added
No just updating dependency and checking that everything still works.

### Screenshots
No UI changes.

### Paired with 
Nobody.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ task wrapper(type: Wrapper) {
 }
 
 allprojects {
-    version = "2.0.2"
+    version = "2.0.3"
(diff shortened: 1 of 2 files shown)


ouchadam:
no conflicts, woohooo

PR #76, opened 2017-07-26
Jelly bean crash

Mecharyry:
## Problem
When trying to play DRM protected content on API level 18 we would see a crash. This occurred because `LocalDrmSessionManager` was targeting API level 19 when we should have been targeting API 18. We wrongly thought that we should be targeting 19 because `ResourceBusyException` is an API 19 exception... but apparently, if you just use `Exception` in your catch it doesn't matter 🤷‍♂️ and yes ExoPlayer does this too in the Demo...

## Solution
Target API level 18 and catch a generic `Exception` rather than the unsupported `Exception`...

### Test(s) added 
No, just changed the target and edited the exception.

### Screenshots
No UI changes.

### Paired with 
@ouchadam and @zegnus, debugging TEAM!

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/LocalDrmSessionManager.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/LocalDrmSessionManager.java
@@ -1,9 +1,6 @@
 package com.novoda.noplayer.internal.exoplayer.drm;
 
 import android.annotation.TargetApi;
-import android.media.MediaCryptoException;
-import android.media.NotProvisionedException;
-import android.media.ResourceBusyException;
 import android.os.Build;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/LocalDrmSessionManager.java
```diff
@@ -38,7 +35,7 @@
         this.handler = handler;
     }
 
-    @TargetApi(Build.VERSION_CODES.KITKAT)
+    @TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR2)
```
Mecharyry: This matches the underlying `MediaDrm` class which is available as of API 18.

Issue #72, opened 2017-07-18
onVideoStopped is double called


Mecharyry:
#### Problem
As per [this](https://github.com/novoda/no-player/blob/master/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java#L310-L319), on `stop` and on `release` both call `onVideoStopped`.

#### Potential Solution
Only emit a state if it is different from the previous state. Similar to how we handle it for the `play` [here](https://github.com/novoda/no-player/blob/master/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java#L123-L128).

#### Impact
Clients should not need to deal with multiple calls to the same state, it could potentially lead to inconsistencies and crashes, especially if we delay patching this.

Mecharyry:
This has been addressed and released as part of v2.0.6

PR #81, opened 2017-08-07
NO-72/Stop double calling state changes

Mecharyry:
## Problem
As per #72, state changes are being double called. An example of this can be seen [here](https://github.com/novoda/no-player/blob/master/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java#L150-L162). Clients do not need to call `stop` as long as they call `release`, `release` calls `stop` internally` which notifies of the `onVideoStopped` event.

## Solution
Swallow double calls of state through the `StateChangedListeners`.

### Test(s) added 
Yes, tested that double calling swallows the most recent event.

### Screenshots
No UI changes.

### Paired with 
Nobody, but I did have a chat with @ouchadam about this :smile:

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/listeners/StateChangedListeners.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/listeners/StateChangedListeners.java
@@ -1,6 +1,7 @@
 package com.novoda.noplayer.internal.listeners;
 
 import com.novoda.noplayer.Player;
+import com.novoda.utils.NoPlayerLog;
 
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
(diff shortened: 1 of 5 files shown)


Mecharyry:
test this please

core/src/main/java/com/novoda/noplayer/internal/listeners/StateChangedListeners.java
```diff
@@ -23,22 +31,40 @@ void clear() {
 
     @Override
     public void onVideoPlaying() {
+        if (currentState == State.PLAYING) {
+            return;
```
ouchadam: do you think it's worth logging these scenarios?

Mecharyry: Yes, can never have too many logs 😄

Issue #82, opened 2017-08-14
Update ExoPlayer


Mecharyry:
#### Problem
ExoPlayer is now on version 2.5.1 and we are still on 2.4.4.

#### Potential Solution
Update to the latest version.

#### Impact
Stability improvements for clients that use the ExoPlayer version of NoPlayer.

Mecharyry:
This has been addressed as part of version 2.0.7

Issue #79, opened 2017-07-27
TODO or NOT_TODO Forwarders


Mecharyry:
#### Problem
We have `TODO` comments in a lot of the `Forwarder` classes. 

#### Potential Solution
We should decide whether we want to forward these events or swallow them.  If swallowing them then we should provide a reason for this as part of the description of the `TODO`.

#### Impact
Provide more information to the `client` applications by forwarding these additional events.
We supply reasoning as to why we are not forwarding these events, increasing confidence in the codebase.

PR #83, opened 2017-08-14
Update ExoPlayer to v2.5.1 

Mecharyry:
## Problem
As per issue #82, we are using an outdated version of `ExoPlayer`.

NoPlayer Version: 2.4.4
Current Version: 2.5.1

## Solution
Use the latest version of `ExoPlayer`.

`ExoPlayer.EventListener` has been deprecated and now exists in `Player.EventListener` with an additional method for `onRepeatModeChanged`.

`DrmSessionManager` has an additional method that checks whether we can acquire a DRM session before attempting to. In the case of `LocalDrmSessionManager` this is a simple assertion that the `DrmInitData` contains a given `DrmScheme`.

### Test(s) added 
Yes, added tests to the previously untested `LocalDrmSessionManager`. Annoyingly, `FrameworkMediaCrypto` is a final class with a package constructor so I have created a `FrameworkMediaCryptoFixture` in order to be able to test the `LocalDrmSessionManager`.

### Screenshots
No UI changes.

### Paired with 
Nobody.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,5 +19,8 @@ buildscript {
 subprojects {
     repositories {
         jcenter()
+        maven {
+            url "https://maven.google.com"
+        }
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ task wrapper(type: Wrapper) {
 }
 
 allprojects {
-    version = "2.0.6"
+    version = "2.0.7"
 }
 
 buildscript {
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/InvalidDrmSession.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/InvalidDrmSession.java
@@ -28,11 +28,6 @@ class InvalidDrmSession implements FrameworkDrmSession {
     }
 
     @Override
-    public boolean requiresSecureDecoderComponent(String mimeType) {
-        throw new IllegalStateException();
-    }
-
-    @Override
     public DrmSessionException getError() {
         return drmSessionException;
     }
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/InvalidDrmSession.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/InvalidDrmSession.java
@@ -47,4 +47,23 @@ class InvalidDrmSession implements FrameworkDrmSession {
     public SessionId getSessionId() {
         return SessionId.absent();
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        InvalidDrmSession that = (InvalidDrmSession) o;
+
+        return drmSessionException != null ? drmSessionException.equals(that.drmSessionException) : that.drmSessionException == null;
+    }
+
+    @Override
+    public int hashCode() {
+        return drmSessionException != null ? drmSessionException.hashCode() : 0;
+    }
 }
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BitrateForwarder.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BitrateForwarder.java
@@ -33,26 +33,26 @@ class BitrateForwarder implements AdaptiveMediaSourceEventListener {
 
     @Override
     public void onLoadStarted(DataSpec dataSpec, int dataType, int trackType, Format trackFormat, int trackSelectionReason, Object trackSelectionData, long mediaStartTimeMs, long mediaEndTimeMs, long elapsedRealtimeMs) {
-        //TODO should we send ?
+        // TODO: should we send?
     }
 
     @Override
     public void onLoadCompleted(DataSpec dataSpec, int dataType, int trackType, Format trackFormat, int trackSelectionReason, Object trackSelectionData, long mediaStartTimeMs, long mediaEndTimeMs, long elapsedRealtimeMs, long loadDurationMs, long bytesLoaded) {
-        //TODO should we send ?
+        // TODO: should we send?
     }
 
     @Override
     public void onLoadCanceled(DataSpec dataSpec, int dataType, int trackType, Format trackFormat, int trackSelectionReason, Object trackSelectionData, long mediaStartTimeMs, long mediaEndTimeMs, long elapsedRealtimeMs, long loadDurationMs, long bytesLoaded) {
-        //TODO should we send ?
+        // TODO: should we send?
     }
 
     @Override
     public void onLoadError(DataSpec dataSpec, int dataType, int trackType, Format trackFormat, int trackSelectionReason, Object trackSelectionData, long mediaStartTimeMs, long mediaEndTimeMs, long elapsedRealtimeMs, long loadDurationMs, long bytesLoaded, IOException error, boolean wasCanceled) {
-        //TODO should we send ?
+        // TODO: should we send?
     }
 
     @Override
     public void onUpstreamDiscarded(int trackType, long mediaStartTimeMs, long mediaEndTimeMs) {
-        //TODO should we send ?
+        // TODO: should we send?
     }
 }
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BufferStateForwarder.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BufferStateForwarder.java
@@ -8,7 +8,7 @@ import com.google.android.exoplayer2.source.TrackGroupArray;
 import com.google.android.exoplayer2.trackselection.TrackSelectionArray;
 import com.novoda.noplayer.Player;
 
-class BufferStateForwarder implements ExoPlayer.EventListener {
+class BufferStateForwarder implements com.google.android.exoplayer2.Player.EventListener {
 
     private final Player.BufferStateListener bufferStateListener;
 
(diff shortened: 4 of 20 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/InvalidDrmSession.java
```diff
@@ -52,4 +47,23 @@ public DrmSessionException getError() {
     public SessionId getSessionId() {
         return SessionId.absent();
     }
+
+    @Override
+    public boolean equals(Object o) {
```
joetimmins: good catch

Mecharyry: It was breaking my tests 😆

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BitrateForwarder.java
```diff
@@ -33,26 +33,26 @@ public void onDownstreamFormatChanged(int trackType, Format trackFormat, int tra
...
     @Override
     public void onLoadError(DataSpec dataSpec, int dataType, int trackType, Format trackFormat, int trackSelectionReason, Object trackSelectionData, long mediaStartTimeMs, long mediaEndTimeMs, long elapsedRealtimeMs, long loadDurationMs, long bytesLoaded, IOException error, boolean wasCanceled) {
-        //TODO should we send ?
+        // TODO: should we send?
     }
 
     @Override
     public void onUpstreamDiscarded(int trackType, long mediaStartTimeMs, long mediaEndTimeMs) {
-        //TODO should we send ?
+        // TODO: should we send?
```
joetimmins: I can see these were really bothering you 👍

Mecharyry: I have a problem 😆

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/EventInfoForwarder.java
```diff
@@ -10,7 +9,7 @@
 
 import java.util.HashMap;
 
-class EventInfoForwarder implements ExoPlayer.EventListener {
+class EventInfoForwarder implements com.google.android.exoplayer2.Player.EventListener {
```
joetimmins: Can we have this interface as an import?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BufferStateForwarder.java
```diff
@@ -26,32 +26,37 @@ public void onPlayerStateChanged(boolean playWhenReady, int playbackState) {
     }
 
     @Override
+    public void onRepeatModeChanged(@com.google.android.exoplayer2.Player.RepeatMode int repeatMode) {
```
joetimmins: We can import this interface no? Also what's with the @ here? 😛

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/EventInfoForwarder.java
Mecharyry: No, annoyingly it can't distinguish between their `Player` interface and ours.

core/src/test/java/com/google/android/exoplayer2/drm/FrameworkMediaCryptoFixture.java
```diff
@@ -0,0 +1,34 @@
+package com.google.android.exoplayer2.drm;
+
+import android.media.MediaCrypto;
+import android.media.MediaCryptoException;
+
+import java.util.UUID;
+
+public final class FrameworkMediaCryptoFixture {
```
joetimmins: 👍

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BufferStateForwarder.java
Mecharyry: This is the retention policy from their code. I thought it was a good idea to keep it because of the `int`.

core/src/test/java/com/novoda/noplayer/internal/exoplayer/drm/LocalDrmSessionManagerTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+        localDrmSessionManager.acquireSession(IGNORED_LOOPER, IGNORED_DRM_DATA);
+
+        ArgumentCaptor<Runnable> argumentCaptor = ArgumentCaptor.forClass(Runnable.class);
+        verify(handler).post(argumentCaptor.capture());
+        argumentCaptor.getValue().run();
+        verify(eventListener).onDrmSessionManagerError(any(DrmSession.DrmSessionException.class));
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.KITKAT)
```
joetimmins: Didn't realise you could use `RequiresApi` in this way, TIL

build.gradle
```diff
@@ -3,7 +3,7 @@ task wrapper(type: Wrapper) {
 }
 
 allprojects {
-    version = "2.0.6"
+    version = "2.0.7"
```
joetimmins: Should we mirror the `exoplayer` versioning and consider this a `2.1.0` release rather than `2.0.7`?

Mecharyry: Could do. I wasn't sure because it has no outward changes for the client. That is unless clients are using something from `ExoPlayer`, which ideally they shouldn't do.

zegnus: What happens if we fix a bug in the no-player library and there is no exo-player update, which version name are we going to have if we align it?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BitrateForwarder.java
zegnus: There is a shortcut to add a date in that TODO, pretty useful:

`type 'todo' and let the code auto-completion do its magic`

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BufferStateForwarder.java
zegnus: just, can we static import the namespace?

zegnus: 💯 for keeping the retention policy

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BitrateForwarder.java
Mecharyry: Yes but the real question is, why would I want to do that 👅

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/EventInfoForwarder.java
```diff
@@ -58,6 +57,15 @@ public void onPlayerStateChanged(boolean playWhenReady, int playbackState) {
     }
 
     @Override
+    public void onRepeatModeChanged(@com.google.android.exoplayer2.Player.RepeatMode int repeatMode) {
+        HashMap<String, String> callingMethodParameters = new HashMap<>();
```
zegnus: Can we do this?

`Map<String, String> callingMethodParameters = new HashMap<>(1);`

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/PlayerOnErrorForwarder.java
```diff
@@ -1,14 +1,13 @@
 package com.novoda.noplayer.internal.exoplayer.forwarder;
 
 import com.google.android.exoplayer2.ExoPlaybackException;
-import com.google.android.exoplayer2.ExoPlayer;
 import com.google.android.exoplayer2.PlaybackParameters;
 import com.google.android.exoplayer2.Timeline;
 import com.google.android.exoplayer2.source.TrackGroupArray;
 import com.google.android.exoplayer2.trackselection.TrackSelectionArray;
 import com.novoda.noplayer.Player;
```
zegnus: Could we rename this class to be `NoPlayer` so that we do not clash with Google's namespace

build.gradle
Mecharyry: I normally use this  http://semver.org/ as a basis for versioning. 

-This hasn't introduced any breaking changes so it is not a major.
-This hasn't introduced any new functionality that is visible to the client. You could argue that an update to ExoPlayer could make this a minor but we haven't provided a way for a client to access these new changes.
- Patch is normally used for bug fixes.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BufferStateForwarder.java
Mecharyry: No, I can't. I don't understand how a static import can work here. The Class name matches in both cases so it won't be able to distinguish between them.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BitrateForwarder.java
zegnus: It adds a standard TODO string already formatted and also gives a clue to the next dev on when that TODO was added

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/PlayerOnErrorForwarder.java
Mecharyry: yea sure, in the next major version of the library 😄  this would be a big breaking change.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/EventInfoForwarder.java
Mecharyry: I'd rather do this in a follow-up PR before releasing https://github.com/novoda/no-player/issues/84

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/BitrateForwarder.java
Mecharyry: I personally do not like `TODO`s anyway. All of these are represented by this issue https://github.com/novoda/no-player/issues/79

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/PlayerOnErrorForwarder.java
zegnus: I'd do it sooner rather than later, it make sense to call it `NoPlayer` anyway ;)

Issue #85, opened 2017-08-22
Use role from manifest to choose audio


Mecharyry:
#### Problem
Currently, if a client wants to select an audio described track they need to have intimate knowledge of the ordering in their manifests to do so. It turns out we don't need to though!

`ExoPlayer` sets `Format.selection` flags based on the `role`. For `main` this will be `1` and for audio described this will be `0`. 

#### Potential Solution
Pull the `Format.selection` flag when creating the `PlayerAudioTrack` and set a `boolean` / `enum` on the object to highlight the track as `audio_described`.

#### Impact
Clients do not need to know the ordering of the manifest.

Mecharyry:
This has been addressed in #86

PR #86, opened 2017-08-22
NP-85/Make AudioDescribed track obvious

Mecharyry:
## Problem
As per issue #85, we need to provide a mechanism that will allow client applications to switch between **main** and **audio described** audio tracks, without needing to know the position from the manifest.

## Solution
Added an `AudioTrackType` to represent the different audio track types, **main** and **audio described**, a **unknown** type has been added to take into account other types that might present themselves in the future. `AudioTrackType` is created passing in a given `format.selectionFlags` which is created based on the role.

The `AudioTrackType` is publically accessible through `PlayerAudioTrack.audioTrackType`.

A first class collection representing a `List<PlayerAudioTrack>` has been added. This includes several support methods:

`containsPlayerAudioTrackWith(AudioTrackType)` returns a boolean representing whether the underlying `List<PlayerAudioTrack>` contains a `PlayerAudioTrack` with the given `AudioTrackType`.

`firstPlayerAudioTrackWith(AudioTrackType)` returns either the first `PlayerAudioTrack` that contains the given `AudioTrackType`, wrapped in an `Optional` or `Optional.absent`.

### Test(s) added 
Yes, tested all of the newly added classes. Added a `Fixture` for the `PlayerAudioTrack` because it was being created in quite a few tests.

### Screenshots
We now have the type shown in the audio selection.

audio_type

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/Player.java
+++ b/core/src/main/java/com/novoda/noplayer/Player.java
@@ -102,7 +102,7 @@ public interface Player extends PlayerState {
     void detach(PlayerView playerView);
 
     /**
-     * Retrieves all of the available {@link PlayerAudioTrack} of a prepared Player as a first class collection.
+     * Retrieves all of the available {@link PlayerAudioTrack} of a prepared Player.
      *
      * @return {@link AudioTracks} that contains a list of available {@link PlayerAudioTrack}.
      * @throws IllegalStateException - if called before {@link Player#loadVideo(Uri, ContentType)}.
--- a/core/src/main/java/com/novoda/noplayer/Player.java
+++ b/core/src/main/java/com/novoda/noplayer/Player.java
@@ -2,6 +2,7 @@ package com.novoda.noplayer;
 
 import android.net.Uri;
 
+import com.novoda.noplayer.model.AudioTracks;
 import com.novoda.noplayer.model.Bitrate;
 import com.novoda.noplayer.model.PlayerAudioTrack;
 import com.novoda.noplayer.model.PlayerSubtitleTrack;
@@ -101,13 +102,13 @@ public interface Player extends PlayerState {
     void detach(PlayerView playerView);
 
     /**
-     * Retrieves all of the available {@link PlayerAudioTrack} of a prepared Player.
+     * Retrieves all of the available {@link PlayerAudioTrack} of a prepared Player as a first class collection.
      *
-     * @return A list of available {@link PlayerAudioTrack}.
+     * @return {@link AudioTracks} that contains a list of available {@link PlayerAudioTrack}.
      * @throws IllegalStateException - if called before {@link Player#loadVideo(Uri, ContentType)}.
      * @see Player.PreparedListener
      */
-    List<PlayerAudioTrack> getAudioTracks() throws IllegalStateException;
+    AudioTracks getAudioTracks() throws IllegalStateException;
 
     /**
      * Selects a given {@link PlayerAudioTrack}.
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
@@ -1,0 +1,22 @@
+package com.novoda.noplayer.internal.exoplayer.mediasource;
+
+public enum AudioTrackType {
+    MAIN(1),
+    AUDIO_DESCRIBED(0),
+    UNKNOWN(-1);
+
+    private final int selectionFlag;
+
+    AudioTrackType(int selectionFlag) {
+        this.selectionFlag = selectionFlag;
+    }
+
+    static AudioTrackType from(int selectionFlag) {
+        for (AudioTrackType audioTrackType : AudioTrackType.values()) {
+            if (audioTrackType.selectionFlag == selectionFlag) {
+                return audioTrackType;
+            }
+        }
+        return UNKNOWN;
+    }
+}
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
@@ -1,8 +1,8 @@
 package com.novoda.noplayer.internal.exoplayer.mediasource;
 
 public enum AudioTrackType {
-    MAIN(1),
-    AUDIO_DESCRIBED(0),
+    MAIN(0),
+    AUDIO_DESCRIBED(1),
     UNKNOWN(-1);
 
     private final int selectionFlag;
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
@@ -2,7 +2,7 @@ package com.novoda.noplayer.internal.exoplayer.mediasource;
 
 public enum AudioTrackType {
     MAIN(0),
-    AUDIO_DESCRIBED(1),
+    ALTERNATIVE(1),
     UNKNOWN(-1);
 
     private final int selectionFlag;
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
@@ -1,6 +1,7 @@
 package com.novoda.noplayer.internal.exoplayer.mediasource;
 
 public enum AudioTrackType {
+
     MAIN(0),
     ALTERNATIVE(1),
     UNKNOWN(-1);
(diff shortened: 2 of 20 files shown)


core/src/main/java/com/novoda/noplayer/model/AudioTracks.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public static AudioTracks from(List<PlayerAudioTrack> audioTracks) {
+        return new AudioTracks(Collections.unmodifiableList(audioTracks));
+    }
+
+    private AudioTracks(List<PlayerAudioTrack> audioTracks) {
+        this.audioTracks = audioTracks;
+    }
+
+    public boolean containsPlayerAudioTrackWith(AudioTrackType trackType) {
```
Mecharyry: I don't create many first class collections and was unsure whether putting these here was a good idea. I would be interested in hearing other peoples thoughts on this. 

As a side note, these methods are fully tested 😄

core/src/main/java/com/novoda/noplayer/model/AudioTracks.java
```diff
@@ -0,0 +1,78 @@
...
+        for (PlayerAudioTrack audioTrack : audioTracks) {
+            if (audioTrack.audioTrackType() == trackType) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public Optional<PlayerAudioTrack> firstPlayerAudioTrackWith(AudioTrackType audioTrackType) {
```
Mecharyry: I dislike returning `null` and we using the `Optional` will allow clients to do `playerAudioTrack.or(defaultCase)`.

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -76,25 +78,32 @@ public void onClick(View v) {
...
                     }).create();
             audioSelectionDialog.show();
         }
 
-        private List<String> mapAudioTrackToLabel(List<PlayerAudioTrack> audioTracks) {
+        private List<String> mapAudioTrackToLabel(AudioTracks audioTracks) {
             List<String> labels = new ArrayList<>();
             for (PlayerAudioTrack audioTrack : audioTracks) {
-                labels.add("Group: " + audioTrack.groupIndex() + " Format: " + audioTrack.formatIndex());
+                String label = String.format(
```
Mecharyry: Showing the type in the label, we no longer need to guess!  🎉

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
```diff
@@ -0,0 +1,22 @@
+package com.novoda.noplayer.internal.exoplayer.mediasource;
+
+public enum AudioTrackType {
+    MAIN(0),
+    AUDIO_DESCRIBED(1),
```
ouchadam: I'm not sure if we should be calling this `AudioDescribed` whilst the most common usecase is audio description it's not enforced by the dash spec 

Annoyingly I can't find the official DASH spec website any more only this which isn't great

http://dashif.org/software/

core/src/main/java/com/novoda/noplayer/Player.java
```diff
@@ -101,13 +102,13 @@
     void detach(PlayerView playerView);
 
     /**
-     * Retrieves all of the available {@link PlayerAudioTrack} of a prepared Player.
+     * Retrieves all of the available {@link PlayerAudioTrack} of a prepared Player as a first class collection.
```
ataulm: not sure this change is needed

core/src/main/java/com/novoda/noplayer/Player.java
```diff
@@ -101,13 +102,13 @@
...
      *
-     * @return A list of available {@link PlayerAudioTrack}.
+     * @return {@link AudioTracks} that contains a list of available {@link PlayerAudioTrack}.
      * @throws IllegalStateException - if called before {@link Player#loadVideo(Uri, ContentType)}.
      * @see Player.PreparedListener
      */
-    List<PlayerAudioTrack> getAudioTracks() throws IllegalStateException;
+    AudioTracks getAudioTracks() throws IllegalStateException;
```
ataulm: `PlayerAudioTracks getPlayerAudioTracks()`

I mean Ideally `PlayerAudioTrack` -> `AudioTrack` 😁 but I guess not possible

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
```diff
@@ -0,0 +1,22 @@
+package com.novoda.noplayer.internal.exoplayer.mediasource;
+
+public enum AudioTrackType {
+    MAIN(0),
```
ataulm: linebreak above?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
```diff
@@ -0,0 +1,22 @@
...
+    AUDIO_DESCRIBED(1),
+    UNKNOWN(-1);
+
+    private final int selectionFlag;
```
ataulm: :+1:

core/src/main/java/com/novoda/noplayer/model/AudioTracks.java
```diff
@@ -0,0 +1,78 @@
...
+        for (PlayerAudioTrack audioTrack : audioTracks) {
+            if (audioTrack.audioTrackType() == audioTrackType) {
+                return Optional.of(audioTrack);
+            }
+        }
+        return Optional.absent();
+    }
+
+    public PlayerAudioTrack getPlayerAudioTrackAt(int position) {
```
ataulm: `get(int position)` is idiomatic enough for a collection type? It'll be `audioTracks.get(i)` vs `audioTracks.getPlayerAudioTrackAt(i)`

core/src/main/java/com/novoda/noplayer/model/AudioTracks.java
ataulm: this method I'm unsure about. Is Optional exposed already in the NoPlayer api (i.e. is it part of any method return value/parameters)?

you already have `containsPlayerAudioTrackWith(AudioTrackType)` - this method could be implemented client side if they wanted it, I would ditch for now.

core/src/main/java/com/novoda/noplayer/model/AudioTracks.java
ataulm: this _is_ the sort of thing you could add to a first class collection but not sure if it'd be better to wait to see if there's a need across projects? Easy to add later but adding it now would mean we're stuck and breaking change to remove it later.

`containsTrackWith(AudioTrackType)` for brevity?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/AudioTrackType.java
ouchadam: if memory serves me correctly this was called `alternative`

core/src/main/java/com/novoda/noplayer/model/AudioTracks.java
Mecharyry: I added this **because** it made it easier and less verbose on a client project 😄 

I'll change the name of all to remove the `PlayerAudioTrack`

core/src/main/java/com/novoda/noplayer/model/AudioTracks.java
Mecharyry: Happy to ditch it 😄

ataulm: yeah ditch this one :D

core/src/main/java/com/novoda/noplayer/model/AudioTracks.java
zegnus: `contains(Type ...)` would be more than enough ;) I would also keep it

ataulm: not `contains(Type)` because this is a collection of `AudioTrack` not `AudioType`.

we discussed also offline that if a manifest contains multiple languages, that there will be multiple tracks with the same type - e.g. "english - main" and "french - main" at which point `containsTrackWith(AudioTrackType)` is not enough information?

zegnus: you can have `contains(TypeA ...)` with as many types as you want; `contains(TypeB ...)`, you don't need to repeat the type in the name of the method, only when using primitives if necessary

ataulm: nope I follow that it would work technically, I meant it's not semantically correct (which is important for first class collection which is all just semantics) for a collection of type X to return true for a method `contains(TypeY)` because it doesn't contain TypeY, it contains things that may be of Type Y.

zegnus: Ah gotcha!

ouchadam:
LETS DO THIS 🛥

PR #94, opened 2017-10-06
Avoid emitting multiple `onComplete` events

Mecharyry:
## Problem
It's possible for client to receive two `onComplete` events from `no-player`. One comes from a `handler` in `exo-player` and another comes `onPause` of the activity. The `onPause` is problematic because internally `exo-player` checks the last state a notifies the `StateChangedListener` which will return `complete` again 😬 

## Solution
Add a `boolean` flag 😷  to the `OnCompletionListeners`. We have this in the `OnPreparedListeners` also but this is not the ideal solution in both cases. It would be better to simplify the logic around the listeners and forwarders in `no-player`. We have created an issue for this at #95 

### Test(s) added 
Added tests for the `CompletionListeners`.

### Screenshots
No UI changes.

### Paired with 
@jonreeve

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/listeners/CompletionListeners.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/listeners/CompletionListeners.java
@@ -9,6 +9,8 @@ class CompletionListeners implements NoPlayer.CompletionListener {
 
     private final Set<NoPlayer.CompletionListener> listeners = new CopyOnWriteArraySet<>();
 
+    private boolean hasCompleted = false;
+
     void add(NoPlayer.CompletionListener listener) {
         listeners.add(listener);
     }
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/novoda/noplayer/internal/listeners/CompletionListeners.java
```diff
@@ -23,8 +25,15 @@ void clear() {
 
     @Override
     public void onCompletion() {
-        for (NoPlayer.CompletionListener listener : listeners) {
-            listener.onCompletion();
+        if (!hasCompleted) {
```
zegnus: you could totally fast return here ;)

core/src/main/java/com/novoda/noplayer/internal/listeners/PreparedListeners.java
```diff
@@ -26,7 +26,7 @@ void clear() {
 
     @Override
     public void onPrepared(PlayerState playerState) {
-        if (hasNotPreviouslyPrepared()) {
+        if (!hasPrepared) {
```
zegnus: same here

PR #96, opened 2017-10-13
Handler seek exception

Mecharyry:
## Problem
We've seen this `stack trace` on some client applications (replicated in no-player)

screen shot 2017-10-13 at 11 33 48

In some client applications we attempt to catch any `IllegalStateExceptions` that occur when calling methods out of sync on `NoPlayer`. Unfortunately, the calling code from this trace does not come from any client app but internally from inside `no-player` because of a workaround that is implemented on `MediaPlayer`:

```
    /**
     * Workaround to fix some devices (nexus 7 2013 in particular) from natively crashing the mediaplayer
     * by starting the mediaplayer before seeking it.
     */
```
See [here](https://github.com/novoda/no-player/blob/master/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java#L151-L160) for the problematic code.

This workaround internally uses a `Handler.postDelayed`. This runnable is never removed on `reset` of `Player` so if a user leaves a `Player` activity before this has executed it will crash. It crashes because on exit of a `Player` activity will release all internal collaborators of `MediaPlayer` which essentially means `mediaPlayer` will be `null` and so failing the `assertIsInPlaybackState()` which will throw an `IllegalStateException` as designed.

## Solution
To find this I increased the `DELAY` on the `Runnable` in the `MediaPlayer` and ran in a client application, navigating away from the activity before this `Runnable` is called = insta crash 😄 

Looking at the code you could add a check for `mediaPlayerFacade.isPlaying` [here](https://github.com/novoda/no-player/blob/master/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java#L151-L160) which will internally do the same checks as are done for the `assertIsInPlaybackState()` but instead just return in the runnable, avoiding the throw that is called later. BUT this seemed like a workaround rather than an actual fix. 

So I added `handler.removeCallbacksAndMessages(null);` which will remove all callbacks and messages whenever `reset()` is called in `MediaPlayerImpl`. For all of those that are worried about the `null` on this method this is what the docs say about this call:

``` 
    /**
     * Remove any pending posts of callbacks and sent messages whose
     * <var>obj</var> is <var>token</var>.  If <var>token</var> is null,
     * all callbacks and messages will be removed.
     */
```

### Test(s) added 
Updated the `MediaPlayerImpl` tests to take into account the reset on the handler.

### Screenshots
No UI changes.

### Paired with 
Nobody.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ task wrapper(type: Wrapper) {
 }
 
 allprojects {
-    version = "3.0.2"
+    version = "3.0.3"
 }
 
 buildscript {
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ task wrapper(type: Wrapper) {
 }
 
 allprojects {
-    version = "3.0.3"
+    version = "3.0.2"
 }
 
 buildscript {
--- a/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
@@ -29,7 +29,7 @@ import java.util.List;
 class AndroidMediaPlayerImpl implements NoPlayer {
 
     private static final VideoPosition NO_SEEK_TO_POSITION = VideoPosition.INVALID;
-    private static final int INITIAL_PLAY_SEEK_DELAY_IN_MILLIS = 500;
+    private static final long INITIAL_PLAY_SEEK_DELAY_IN_MILLIS = 500;
 
     private final List<SurfaceHolderRequester.Callback> surfaceHolderRequesterCallbacks = new ArrayList<>();
 
--- a/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
@@ -2,7 +2,6 @@ package com.novoda.noplayer.internal.mediaplayer;
 
 import android.media.MediaPlayer;
 import android.net.Uri;
-import android.os.Handler;
 import android.view.SurfaceHolder;
 import android.view.View;
 
@@ -38,7 +37,7 @@ class AndroidMediaPlayerImpl implements NoPlayer {
     private final AndroidMediaPlayerFacade mediaPlayer;
     private final MediaPlayerForwarder forwarder;
     private final CheckBufferHeartbeatCallback bufferHeartbeatCallback;
-    private final Handler handler;
+    private final DelayedActionExecutor delayedActionExecutor;
     private final Heart heart;
     private final PlayerListenersHolder listenersHolder;
     private final LoadTimeout loadTimeout;
@@ -59,7 +58,7 @@ class AndroidMediaPlayerImpl implements NoPlayer {
                            CheckBufferHeartbeatCallback bufferHeartbeatCallback,
                            LoadTimeout loadTimeout,
                            Heart heart,
-                           Handler handler,
+                           DelayedActionExecutor delayedActionExecutor,
                            BuggyVideoDriverPreventer buggyVideoDriverPreventer) {
         this.mediaPlayerInformation = mediaPlayerInformation;
         this.mediaPlayer = mediaPlayer;
@@ -68,7 +67,7 @@ class AndroidMediaPlayerImpl implements NoPlayer {
         this.bufferHeartbeatCallback = bufferHeartbeatCallback;
         this.loadTimeout = loadTimeout;
         this.heart = heart;
-        this.handler = handler;
+        this.delayedActionExecutor = delayedActionExecutor;
         this.buggyVideoDriverPreventer = buggyVideoDriverPreventer;
     }
 
@@ -151,9 +150,9 @@ class AndroidMediaPlayerImpl implements NoPlayer {
     private void initialSeekWorkaround(SurfaceHolder surfaceHolder, final VideoPosition initialPlayPosition) throws IllegalStateException {
         listenersHolder.getBufferStateListeners().onBufferStarted();
         initialisePlaybackForSeeking(surfaceHolder);
-        handler.postDelayed(new Runnable() {
+        delayedActionExecutor.performAfterDelay(new DelayedActionExecutor.Action() {
             @Override
-            public void run() {
+            public void perform() {
                 seekWithIntentToPlay(initialPlayPosition);
             }
         }, INITIAL_PLAY_SEEK_DELAY_IN_MILLIS);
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -3,7 +3,7 @@ task wrapper(type: Wrapper) {
 }
 
 allprojects {
-    version = "3.0.2"
+    version = "3.0.3"
```
ataulm: this should be done separately to make it easier to revert PRs

joetimmins: yes it should

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -333,6 +333,7 @@ public void release() {
     }
 
     private void reset() {
+        handler.removeCallbacksAndMessages(null);
```
lgvalle: @Mecharyry Isn't this risky? If more runnables are posted to that handler in the future you'll cancel all of them here without explicitly saying so. Maybe that's desirable but maybe is not.
There's also `handler.removeCallbacks(runnable)` where you need to specify the runnable you want to kill

ouchadam: 👍 would be safer to keep a reference to the runnable and cancel that instead _orrrrrr_ create a new handler which is only used for this workaround, then we should be good to nuke it.

Mecharyry: I did consider this. Ultimately doing `handler.removeCallbacks(runnable)` would require me to keep track of all `runnables` that this class can produce. Really this class should never have more than one runnable in play at any one time, the delay is short and there are a few conditionals that would need to be triggered to make this happen. 

Perhaps the biggest reason for not using `handler.removeCallbacks(runnable)` is the purpose of the `reset`. `reset` is used to restore the `MediaPlayerImpl` to it's default state, that is, the state it contains immediately after creation. In creation the `MediaPlayerImpl` does not contain a `handler` with `runnables` and therefore it shouldn't after a `reset` is called. 

Thanks for mentioning it ❤️  it's nice to explain this to put it in my head more 😄

Mecharyry:
Going to close this for now so I can deal with the concerns around the handler.

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
Mecharyry: I'll opt to rename and potentially wrap this handler to make this more explicit, would prefer over creating new handlers or keeping track of runnables.

Mecharyry:
OK! So full disclosure, turns out I was horribly wrong about that `handler` 😆  it is being used by other collaborators, they are just created outside in the `NoPlayerMediaPlayerCreator` and I didn't see them the first time! So thanks @lgvalle 

What this means is that we can't just remove all the callbacks because this might screw up the other collaborators. Now we are tracking the runnables and removing the ones we are interested in when the `reset` method is called 😄

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/DelayedActionExecutor.java
```diff
@@ -0,0 +1,43 @@
+package com.novoda.noplayer.internal.mediaplayer;
+
+import android.os.Handler;
+
+import java.util.Iterator;
+import java.util.Map;
+
+class DelayedActionExecutor {
```
Mecharyry: Tracks all of the `Runnable`s for the workaround in `mediaPlayer` and removes all `Runnable`s when `reset` is called in `mediaPlayer`.

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/NoPlayerMediaPlayerCreator.java
```diff
@@ -46,6 +48,7 @@ public AndroidMediaPlayerImpl create(Context context) {
             CheckBufferHeartbeatCallback bufferHeartbeatCallback = new CheckBufferHeartbeatCallback();
             Heart heart = Heart.newInstance(handler);
             MediaPlayerTypeReader mediaPlayerTypeReader = new MediaPlayerTypeReader(new SystemProperties(), Build.VERSION.SDK_INT);
+            DelayedActionExecutor delayedActionExecutor = new DelayedActionExecutor(handler, new HashMap<DelayedActionExecutor.Action, Runnable>());
```
Mecharyry: This is an internal creator so I think it is valid to pass the map through the constructor here.

core/src/test/java/com/novoda/noplayer/internal/mediaplayer/DelayedActionExecutorTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static org.fest.assertions.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.BDDMockito.willAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+public class DelayedActionExecutorTest {
```
Mecharyry: Unit tests making sure that the map and handler are properly handled.

core/src/test/java/com/novoda/noplayer/internal/mediaplayer/DelayedActionExecutorTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private static final long ANY_DELAY_IN_MILLIS = 10;
+
+    private final DelayedActionExecutor.Action action = mock(DelayedActionExecutor.Action.class);
```
Mecharyry: We've had this conversation on another project recently about the performance impact of using the annotation so I decided to do it here. Hopefully we will edit our tests in future to reflect this!

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerImpl.java
```diff
@@ -30,15 +29,15 @@
 class AndroidMediaPlayerImpl implements NoPlayer {
 
     private static final VideoPosition NO_SEEK_TO_POSITION = VideoPosition.INVALID;
-    private static final int INITIAL_PLAY_SEEK_DELAY_IN_MILLIS = 500;
+    private static final long INITIAL_PLAY_SEEK_DELAY_IN_MILLIS = 500;
```
Mecharyry: `Handler` and `Executor` both take long so now this is one too.

PR #98, opened 2017-10-24
PlayerVideoTrack

Mecharyry:
## Problem
We want to be able to display an overlay for the currently selected track and to know all of the available tracks.

## Solution
Implement a `TrackSelector` similar to what was done for [PlayerAudioTrack](https://github.com/novoda/no-player/blob/master/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/ExoPlayerAudioTrackSelector.java). Expose two new methods on `NoPlayer`:

- getSelectedVideoTrack
- getVideoTracks

### Test(s) added 
Yes, updated the `Facade` and `TrackSelector` tests.

### Screenshots
Overlay was added to show potential usage, this has not been added to no-player.
no-player-video-info mp4

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
@@ -140,8 +140,15 @@ class ExoPlayerFacade {
 
     PlayerVideoTrack getSelectedVideoTrack() {
         assertVideoLoaded();
-        Format videoFormat = exoPlayer.getVideoFormat();
-        return new PlayerVideoTrack(contentType, videoFormat.width, videoFormat.height, (int) videoFormat.frameRate, videoFormat.bitrate);
+        Format format = exoPlayer.getVideoFormat();
+        return new PlayerVideoTrack(
+                format.id,
+                contentType,
+                format.width,
+                format.height,
+                (int) format.frameRate,
+                format.bitrate
+        );
     }
 
     List<PlayerVideoTrack> getVideoTracks() {
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
@@ -4,6 +4,7 @@ import android.net.Uri;
 import android.support.annotation.Nullable;
 import android.view.SurfaceHolder;
 
+import com.google.android.exoplayer2.Format;
 import com.google.android.exoplayer2.SimpleExoPlayer;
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -29,22 +32,27 @@
     private final MediaSourceFactory mediaSourceFactory;
     private final ExoPlayerAudioTrackSelector audioTrackSelector;
     private final ExoPlayerSubtitleTrackSelector subtitleTrackSelector;
+    private ExoPlayerVideoTrackSelector exoPlayerVideoTrackSelector;
```
zegnus: if isn't final, can you move it after the final ones?

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -38,6 +40,7 @@
     private int currentBufferPercentage;
 
     private MediaPlayerCreator mediaPlayerCreator;
+    private ContentType contentType;
```
zegnus: Can this be `@Nullable`? and if can, shall we protect methods in the class for that case?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
Mecharyry: I think my IDE settings are messed up, this should have happened automatically when I extracted the field 😢

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
Mecharyry: Good point, the `MediaPlayer` should be too 🤔

Mecharyry:
@zegnus the demo is not updated with this example, the `PlayerVideoTrack` is made available to our clients, who can do this. I didn't think it was a necessary part of the demo, this might change in the future.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
Mecharyry: b8c8f48

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
Mecharyry: dc6ee27

PR #102, opened 2017-11-06
Clear track selections

Mecharyry:
## Problem
We want to be able to provide more debugging capabilities in our client applications that use `no-player` such as selecting a video rendition. 

## Solution
Use the same mechanism that exists for hiding `subtitles` but apply it to `audio` and `video` tracks. 

### Screenshots
The connection is too good here so if I use auto it automatically goes to the highest rendition after the set buffer time (internal) 😆 
device-2017-11-06-122227

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/NoPlayer.java
+++ b/core/src/main/java/com/novoda/noplayer/NoPlayer.java
@@ -104,6 +104,15 @@ public interface NoPlayer extends PlayerState {
     void detach(PlayerView playerView);
 
     /**
+     * Retrieves all of the available {@link PlayerVideoTrack} of a prepared Player.
+     *
+     * @return a list of available {@link PlayerVideoTrack} of a prepared Player.
+     * @throws IllegalStateException - if called before {@link NoPlayer#loadVideo(Uri, ContentType)}.
+     * @see NoPlayer.PreparedListener
+     */
+    List<PlayerVideoTrack> getVideoTracks() throws IllegalStateException;
+
+    /**
      * Selects a given {@link PlayerVideoTrack}.
      *
      * @param videoTrack the video track to select.
@@ -123,13 +132,12 @@ public interface NoPlayer extends PlayerState {
     Optional<PlayerVideoTrack> getSelectedVideoTrack() throws IllegalStateException;
 
     /**
-     * Retrieves all of the available {@link PlayerVideoTrack} of a prepared Player.
+     * Clears the {@link PlayerVideoTrack} selection made in {@link NoPlayer#selectVideoTrack(PlayerVideoTrack)}.
      *
-     * @return a list of available {@link PlayerVideoTrack} of a prepared Player.
+     * @return whether the clear was successful.
      * @throws IllegalStateException - if called before {@link NoPlayer#loadVideo(Uri, ContentType)}.
-     * @see NoPlayer.PreparedListener
      */
-    List<PlayerVideoTrack> getVideoTracks() throws IllegalStateException;
+    boolean clearVideoTrackSelection() throws IllegalStateException;
 
     /**
      * Retrieves all of the available {@link PlayerAudioTrack} of a prepared Player.
@@ -167,7 +175,7 @@ public interface NoPlayer extends PlayerState {
     List<PlayerSubtitleTrack> getSubtitleTracks() throws IllegalStateException;
 
     /**
-     * Selects a given {@link PlayerSubtitleTrack} on an attached PlayerView.
+     * Selects and shows a given {@link PlayerSubtitleTrack} on an attached PlayerView.
      *
      * @param subtitleTrack the subtitle track to select.
      * @return whether the selection was successful.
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/novoda/noplayer/NoPlayer.java
```diff
@@ -170,9 +186,10 @@
     /**
      * Clear and hide the subtitles on an attached PlayerView.
      *
+     * @return whether the hide was successful.
      * @throws IllegalStateException - if called before {@link NoPlayer#loadVideo(Uri, ContentType)}.
      */
-    void hideSubtitleTrack() throws IllegalStateException;
+    boolean hideSubtitleTrack() throws IllegalStateException;
```
Mecharyry: Changed this to return `boolean` to be consistent with the other clear methods introduced.

Mecharyry:
❌  DO NOT MERGE ❌  I want to try a local release into a client application before we merge this.

core/src/main/java/com/novoda/noplayer/NoPlayer.java
```diff
@@ -132,13 +132,21 @@
...
 
     /**
-     * Retrieves all of the available {@link PlayerVideoTrack} of a prepared Player.
+     * Clears the {@link PlayerVideoTrack} selection made in {@link NoPlayer#selectVideoTrack(PlayerVideoTrack)}.
      *
-     * @return a list of available {@link PlayerVideoTrack} of a prepared Player.
+     * @return whether the clear was successful.
+     * @throws IllegalStateException - if called before {@link NoPlayer#loadVideo(Uri, ContentType)}.
+     */
+    boolean clearVideoTrackSelection() throws IllegalStateException;
```
ataulm: is there thought to whether we'll remove thrown exceptions on APIs where we return boolean to indicate success?

Mecharyry: Good point. The `IllegalStateException` is used when a client calls any of these methods before the `player` is prepared, I guess we could do away with the throws in these cases but then the reasoning as to why the method returned false would be hidden. I guess you could argue the reasoning would be hidden anyway, so why return a `boolean` at all 😄  

I think you're right, this needs more thought 👅

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/ExoPlayerTrackSelector.java
```diff
@@ -27,10 +27,14 @@ TrackGroupArray trackGroups(TrackType trackType, RendererTypeRequester rendererT
         return audioRendererIndex.isAbsent() ? TrackGroupArray.EMPTY : trackInfo().getTrackGroups(audioRendererIndex.get());
     }
 
-    void clearSelectionOverrideFor(TrackType trackType, RendererTypeRequester rendererTypeRequester) {
-        Optional<Integer> audioRendererIndex = rendererTrackIndexExtractor.extract(trackType, mappedTrackInfoLength(), rendererTypeRequester);
-        if (audioRendererIndex.isPresent()) {
-            trackSelector.clearSelectionOverrides(audioRendererIndex.get());
+    boolean clearSelectionOverrideFor(TrackType trackType,
+                                      RendererTypeRequester rendererTypeRequester) {
```
ataulm: rando line break?

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -85,24 +88,34 @@ public void onClick(View v) {
...
             adapter.addAll(mapVideoTrackToLabel(videoTracks));
             AlertDialog videoTrackSelectionDialog = new AlertDialog.Builder(MainActivity.this)
                     .setTitle("Select Video track")
                     .setAdapter(adapter, new DialogInterface.OnClickListener() {
                         @Override
                         public void onClick(DialogInterface dialog, int position) {
-                            PlayerVideoTrack videoTrack = videoTracks.get(position);
-                            player.selectVideoTrack(videoTrack);
+                            switch (position) {
+                                case 0:
```
ataulm: should this be an if else?

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -124,21 +137,28 @@ private void showAudioSelectionDialog() {
                     .setAdapter(adapter, new DialogInterface.OnClickListener() {
                         @Override
                         public void onClick(DialogInterface dialog, int position) {
-                            PlayerAudioTrack audioTrack = audioTracks.get(position);
-                            player.selectAudioTrack(audioTrack);
+                            switch (position) {
```
ataulm: same - if/else?

demo/src/main/java/com/novoda/demo/MainActivity.java
Mecharyry: Yea it should 😆  this is what I get for reusing some of our code.

core/src/main/java/com/novoda/noplayer/NoPlayer.java
zegnus: we need a trilean `true, false, error` 💃

Issue #103, opened 2017-11-09
Avoid creating objects for every second of video


jonreeve:
#### Problem

Strong types are definitely a generally good thing, but in some cases not so appropriate. For VideoPosition we are creating one any time someone queries it. The heartbeat callback is used by many listeners who then query the position, every second. Creating these objects is definitely overkill and could even lead to heap fragmentation, etc. It's quite typical in video work to simply use a 'long seconds' for this kind of thing.

#### Potential Solution

Use a simple long for position.

Worth considering updating the heartbeat callback method signature to simply pass the position and duration (as longs) at this point, since every listener generally just asks for these two things. The player reference shouldn't really be needed, and if it really is needed then when the listener is created it probably could have been given that reference directly.

#### Impact

* Changes public API
* Avoids a lot of unnecessary object churn.

ouchadam:
seems like this has been addressed  👍

Issue #101, opened 2017-11-01
Explicitly separate VideoPosition and VideoDuration concepts


ataulm:
#### Problem

The `VideoPosition` type is being misused or the method names are inaccurate.

```java
    public VideoPosition minus(VideoPosition other)
    public VideoPosition plus(VideoPosition other)
```

We can like VideoPosition to a point in time, like "3pm" or "Tuesday". You can get the _difference_ between two points in time, but it should result in an absolute value (non-negative).

`VideoPosition` is currently being used in two ways:

- a point in time relative to the start of the video
- a length of time

#### Potential Solution

Introduce a new type for `TimeLength` or repurpose `VideoDuration` to represent _any_ length of time (this is the smaller change since at the moment `VideoDuration` represents the _full_ length of a video).

```java
    public VideoPosition minus(VideoDuration other)
    public VideoPosition plus(VideoDuration other)
```
#### Impact

The impact of the types being used indiscriminately is confusion/ambiguity in code. For example:

```java
private static final VideoPosition CREDITS_POSITION_FROM_END_OF_VIDEO = VideoPosition.fromSeconds(60);

private VideoPosition getApproximatePositionOfCreditsGivenFull(VideoDuration videoDuration) {
    VideoPosition endPosition = VideoPosition.fromMillis(videoDuration.inMillis());
    return endPosition.minus(CREDITS_POSITION_FROM_END_OF_VIDEO);
}
```

I've seen (something like) this in production where the constant isn't accurate. If `VideoPosition` is a point in the video defined as millis from the _start_ of the video, then using `VideoPosition` to define a constant of "60 seconds before the end of the video" is impossible without knowing the length of the video.

To do this correctly with the current API we would have:

```java
private static final long CREDITS_POSITION_FROM_END_OF_VIDEO_MILLIS = TimeUnit.SECONDS.toMillis(60);

private VideoPosition getApproximatePositionOfCreditsGivenFull(VideoDuration videoDuration) {
    long creditsPosition = videoDuration.inMillis() - CREDITS_POSITION_FROM_END_OF_VIDEO_MILLIS;
    return VideoPosition.fromMillis(creditsPosition);
}
```
which is fine, except it's not clear from the API that this is the correct thing to do.

ataulm:
what VIdeoPosition/Duration?

PR #106, opened 2017-11-11
NO-103/Replace VideoPosition with long

Mecharyry:
## Problem
We have noticed in client applications a lot of memory churn which seems to be directly related to the creation and destruction of `VideoPosition` and `VideoDuration`, on every tick of the video. 

See #103 for more information.

## Solution
Replace `VideoPosition` with `long` but ensuring all naming associated with `VideoPosition` highlights that this new `long` is based on the `TimeUnit` milliseconds.

### Test(s) added 
Updated all of the associated tests that use `VideoPosition` in some way. 

### Screenshots
No UI changes.

### Paired with 
Nobody.


core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -131,8 +130,8 @@ public void play() throws IllegalStateException {
     }
 
     @Override
-    public void play(VideoPosition position) throws IllegalStateException {
-        seekTo(position);
+    public void play(long positionInMillis) throws IllegalStateException {
```
rock3r: Given we're changing the APIs, can we rename this method to `playAt(long)` or `playFrom(long)`?

Mecharyry: Yes of course 😄  happy to do any interface changes that anyone suggests. This will be a major release given the breaking changes.

ouchadam:
my preference would be to keep the type and have a cache

rock3r:
I would seriously avoid having a cache. It's not going to solve the issue unless you have a humongous cache, and has the cache overhead (you need a lookup on every single event)

ouchadam:
a simple single value cache to avoid all the heartbeat listeners creating a new instance, then see if the big change to long is still required

rock3r:
That makes some more sense than what I had thought when you said cache 😆 But not sure how much of a difference it makes when it's potentially called on every frame. Looks like it needs some measuring and comparison.

Mecharyry:
I'm concerned at getting the time necessary to do this comparison and also having time to implement a cache ready to push to clients ready for Christmas.

ouchadam:
I'm easy, will let yous decide 👍

zegnus:
I'd just use the long and avoid creating extra complexity for the caches...

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -191,14 +191,14 @@ int getDuration() throws IllegalStateException {
         return mediaPlayer.getDuration();
     }
 
-    int getCurrentPosition() throws IllegalStateException {
+    int currentPositionInMillis() throws IllegalStateException {
```
jonreeve: interesting that theirs is an int... I guess that is fairly big in Java. Traditional to use long for video timestamps but I guess in this context int will cover up to almost 600 hours so is sufficient. Should we use the same for consistency, to avoid confusion? Or better to go with more of a typical value for video things?

core/src/main/java/com/novoda/noplayer/model/VideoPosition.java
```diff
@@ -1,118 +0,0 @@
...
-
-    public long inMillis() {
-        return positionInMillis;
-    }
-
-    public int inImpreciseMillis() {
-        return (int) positionInMillis;
-    }
-
-    public int inImpreciseSeconds() {
```
jonreeve: Wondering how much all the methods on this type were used by user(s) of the library? If they're particularly useful we might consider keeping them available as static. This is the one time I'd be willing to suggest a big bundle of static functions as a possibly good thing 😆

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
Mecharyry: I'd go with the more typical, always found it a bit strange that they went with `int` it would work in most cases I guess but 🤷‍♂️

core/src/main/java/com/novoda/noplayer/model/VideoPosition.java
Mecharyry: When I do a local release I'll check to see how much we use these and possibly add some sort of utility class to help out. Although @ataulm did mention that some of these are confusing / incorrect https://github.com/novoda/no-player/issues/101

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -91,9 +90,9 @@ void pause() throws IllegalStateException {
         exoPlayer.setPlayWhenReady(false);
     }
 
-    void seekTo(VideoPosition position) throws IllegalStateException {
+    void seekTo(long positionInMillis) throws IllegalStateException {
```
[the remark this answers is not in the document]
Mecharyry: So an alternative would be to have a cache for the `VideoPosition` on the `heartbeat` callback or to use only the long for the `heartbeat` callback?

core/src/main/java/com/novoda/noplayer/model/VideoPosition.java
ataulm: I think #101 could become obsolete after some of these changes are complete.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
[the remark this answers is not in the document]
jonreeve: No wasn't suggesting all that caching and such. heartbeat defo should not be instantiating types. long timestamps is typical, performant, simple. I was just meaning this looks like something the client calls, not every second, so no objection to using a strong type here if it helps.

ouchadam:
my bad, I forgot to approve 👻

Issue #104, opened 2017-11-10
Deprecate getListeners() in favour of add + remove methods


jonreeve:
#### Problem

Having `player.getListeners().addMyListener(X)` rather than simply `player.addMyListener(X)` makes for a more complicated interface and more brittle tests. Also applies to remove methods.

#### Potential Solution

Add the latter method, and ultimately migrate away from the old one, but we could simply deprecate it for now.

#### Impact

Nice and safe if we just add these and deprecate the old ones. Really good for testing things that use noplayer as a collaborator though.

ouchadam:
We originally had this and chose to avoid as the `NoPlayer` is an interface and meant adding wayyy too many methods to impls.

Mecharyry:
@ouchadam -> https://github.com/novoda/no-player/pull/108

ouchadam:
cc'ing @Dorvaryn and @fourlastor and they were part of the original discussions (if I remember correctly)

Mecharyry:
Won't fix. Client applications can wrap the `listeners` providing a more accessible api if they wish.

PR #108, opened 2017-11-11
NO-104/Remove get listeners

Mecharyry:
## Problem
As per issue #104 we should remove / deprecate `getListeners` from `NoPlayer`, it's making the adding and removing of listeners on the `no-player` unnecessarily complicated.

## Solution
Implement `Listeners` on the `NoPlayer` interface, exo-player and media player, forwarding all additions and removals to the player listeners holder. 

### Paired with 
Nobody.


ouchadam:
I don't see the benefit, we're now forcing implementors of `NoPlayer` to include listener add/removing logic, before this was handled by a single class which the implementor could choose to reuse or customise

Mecharyry:
Just Java docs now 😆  it would be more appropriate for client applications to decide how to interact with the listener holder, whether through `getListeners` or wrapping themselves.

Issue #97, opened 2017-10-24
Remove deprecated TextOutput method


Mecharyry:
#### Problem
`exoPlayer.setTextOutput(textRendererOutput.output());` has been deprecated by ExoPlayer.

#### Potential Solution
Use the new recommended method `#addTextOutput(TextRenderer.Output)` and test in client application.

#### Impact
Deprecated method, no guarantee that it will continue to be supported in the future.

PR #109, opened 2017-11-11
NO-104/Deprecated setTextOutput

Mecharyry:
## Problem
`exoPlayer.setTextOutput(textRendererOutput.output());` has been deprecated on `ExoPlayer` for awhile, we should switch to using the new `addTextOutput`. See #97 for more information.

## Solution
Do what we should and use the non-deprecated method. Keeping a reference to the `TextOutput` in the `ExoPlayerTwoImpl` so that we can remove the exact instance that we need in `detach` and `remove`.

### Paired with 
Nobody.


core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -256,14 +257,21 @@ public boolean showSubtitleTrack(PlayerSubtitleTrack subtitleTrack) throws Illeg
     }
 
     private void setSubtitleRendererOutput() throws IllegalStateException {
-        TextRendererOutput textRendererOutput = new TextRendererOutput(playerView);
+        removeSubtitleRenderer();
+        textRendererOutput = new TextRendererOutput(playerView);
         exoPlayer.setSubtitleRendererOutput(textRendererOutput);
     }
 
+    private void removeSubtitleRenderer() {
```
rock3r: This should be after `hideSubtitleTrack()`

Issue #95, opened 2017-10-06
Simplify NoPlayer Listeners & Forwarders


jonreeve:
#### Problem

There are many duplications around all the types of listener. A listener interface for each, a forwarder for each, a "collection of listeners" for each (that has too much logic in it for just that; some track state, etc). The ultimate interface segregation of the listeners is a good aim, but it doesn't justify all this duplication.

#### Potential Solution

Unify many if not all of these forwarders and listener collections and so on, and simply handle the interface segregation at the outermost layer.

#### Impact

There's a great cognitive load in following all these listeners and forwarders and how they all interact and which one you're viewing at any moment because they all look so similar, and there's changes that have been made to one or two but not all, because of the duplication. Simplifying this need not change the outermost interface to this, but would make following and maintaining this code much easier.

Mecharyry:
Version 2.6.0 of `exo-player` introduces the abstract class `Player.DefaultEventListener` which can be used over the more verbose `Player.EventListener`.

PR #115, opened 2017-12-05
Exo player update

Mecharyry:
## Problem
We are out of date with `exo-player` 😱 

## Solution
Update `no-player` to use the latest version of `exo-player`. https://github.com/google/ExoPlayer/blob/release-v2/RELEASENOTES.md

Most notable from the releases are:
- Workaround broken AAC decoders on Galaxy S6
- Fix memory leak when seeking with repeated periods
- Fix potential IndexOutOfBoundsException when calling ExoPlayer.getDuration
- Fixed adaptive track selection logic for live playbacks

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -3,8 +3,8 @@ apply plugin: 'bintray-release'
 apply plugin: 'jacoco'
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion "25.0.3"
+    compileSdkVersion 27
+    buildToolsVersion "26.0.2"
 
     defaultConfig {
         minSdkVersion 16
@@ -22,7 +22,8 @@ android {
 }
 
 dependencies {
-    compile 'com.google.android.exoplayer:exoplayer:r2.5.4'
+
+    compile 'com.google.android.exoplayer:exoplayer:2.6.0'
     compile 'com.novoda:notils-java:3.0.2'
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:2.7.22'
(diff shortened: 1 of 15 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/OnCompletionForwarder.java
```diff
@@ -28,6 +28,11 @@ public void onRepeatModeChanged(@Player.RepeatMode int repeatMode) {
         // TODO: should we send?
     }
 
+    @Override
+    public void onShuffleModeEnabledChanged(boolean shuffleModeEnabled) {
```
Mecharyry: These listeners are less than ideal but we plan to work on tidying these up as per https://github.com/novoda/no-player/issues/95

core/build.gradle
```diff
@@ -3,8 +3,8 @@ apply plugin: 'bintray-release'
 apply plugin: 'jacoco'
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion "25.0.3"
+    compileSdkVersion 27
+    buildToolsVersion "26.0.2"
```
rock3r: Why not 27.0.1 at this point, given the compile is 27?

demo/build.gradle
```diff
@@ -1,13 +1,13 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion "25.0.3"
+    compileSdkVersion 27
+    buildToolsVersion "26.0.2"
```
rock3r: Ditto here, `27.0.1`

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue Oct 25 13:38:18 BST 2016
+#Tue Dec 05 11:00:47 GMT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.5-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
```
rock3r: Why not 4.3.1?

PR #118, opened 2017-12-06
Add static analysis

rock3r:
## Problem

We're not doing static analysis on this project...

## Solution

Add the standard static analysis setup, with rules. Also changed `compile` dependencies to `implementation` for greater justice. This also includes all the fixes from #119

### Test(s) added 

Nothing to test.

### Paired with 

Nobody


rock3r:
Will be unblocked once #119 is merged into this branch!

PR #119, opened 2017-12-06
Static analysis fixes

rock3r:
## Problem

After #118 we have static analysis, but that highlights a lot of problems :(

## Solution

Address them all! There's as little breaking API and behaviour changes, but there is _some_ minor changes (mostly `NPE`s becoming `ISE`s). This should mean after this PR is shipped **a new major version is required** on the next release!

### Test(s) added 

No, but the existing ones still run a-ok.

### Paired with 

Nobody


rock3r:
test this please

PR #125, opened 2017-12-20
NO-124/Moving internal utils to the internal package

### LINKED ISSUE #124: utils package should move to internal
#### Problem

Currently the `utils` package is top level, we should move to `internal` to highlight these classes aren't meant for client consumption

#### Potential Solution

Move to `internal/utils`

#### Impact

Clients which are already using these classes will have to reimport the package but they will now be aware these classes are internal

ouchadam:
Fixes #124 

## Problem

The internal utils are in a top level package, they should be inside the `internal` package.

## Solution

Move the classes!

### Test(s) added 

nothing to test

### Paired with 
Nobody


hal9002:
Build succeeded

PR #126, opened 2017-12-20
NO-110/Remove notils dependency

### LINKED ISSUE #110: Remove notils dependency
#### Problem

We're including the no-tils library for a single `Exception` 
https://github.com/novoda/no-player/blob/83b2aca946ba4010c255fa8acfed024bc4ff732f/core/src/main/java/com/novoda/noplayer/PlayerType.java#L26

#### Potential Solution

To replace with our own custom exception.

#### Impact

- Decreases maintenance cost of having to keep notils up to date
- If clients were catching this error they will crash

ouchadam:
Fixes #110 

## Problem

We're including [notils](https://github.com/novoda/notils) for the sake of 1 helper exception.

## Solution

Replace the library exception with a custom one. 

### Test(s) added 

To prove the `UnknownPlayerTypeException` is thrown

### Paired with 
Nobody


hal9002:
Build failed

PR #128, opened 2017-12-21
NO-112/Expose repeating playback

### LINKED ISSUE #112: Add api to set repeating playback
#### Problem

We do not currently expose the repeating playback modes of the Media and Exo players

#### Potential Solution

Expose the apis.

ouchadam:
Fixes #112 

## Problem

We're not exposing the repeating playback modes

## Solution

Exposes `player.setRepeating(true/false)` 

### Test(s) added 

Around setting the correct exoplayer repeating mode

### Paired with 
Nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/NoPlayer.java
+++ b/core/src/main/java/com/novoda/noplayer/NoPlayer.java
@@ -200,7 +200,7 @@ public interface NoPlayer extends PlayerState {
     /**
      * Set the Player to repeat the content
      *
-     * @param repeating true to set repeating, false disable
+     * @param repeating true to set repeating, false to disable.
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/novoda/noplayer/NoPlayer.java
```diff
@@ -197,6 +197,14 @@
      */
     boolean hideSubtitleTrack() throws IllegalStateException;
 
+    /**
+     * Set the Player to repeat the content
+     *
+     * @param repeating true to set repeating, false disable
```
Mecharyry: > @param repeating true to set repeating, false to disable.

Don't forget the fullstops 👅

ouchadam: oops! will fix 👍

PR #131, opened 2018-03-06
Polish error handling and mapping

juankysoriano:
## Problem

The error mapping done in `ExoPlayerErrorMapper` was wrong. the method receives a `ExoPlaybackException extends Exception` and immediately checks for it being an instance of particular exceptions. This is wrong as that check will always evaluate to false given that `ExoPlaybackException` is a wrapping exception.

In addition to that it would be good to have a fine granularity on the type of exception in the case it is unknown so client can elaborate better on error handling.

Finally we were prematurely invoking `NoPlayer.ErrorListener.onPlalyerError` for two potential errors. `Media.onLoadError` and `Drm.onDrmSessionManagerError` for client handling. According to the documentation we should only handle those cases for logging.

## Solution

For the mapping we should instead check the type of `ExoPlayerErrorMapper` and elaborate accordingly. 

For the premature invocations honor the documentation by just doing logging about that

### Test(s) added 

Yes, to ensure the `ErrorMapper` does the job as expected

### Screenshots

No UI changes

### Paired with 

Nobody, but chatted a bit with @Mecharyry


core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/DrmSessionErrorForwarder.java
```diff
@@ -1,38 +0,0 @@
...
-        this.errorListener = errorListener;
-    }
-
-    @Override
-    public void onDrmKeysLoaded() {
-        // TODO: should we send?
-    }
-
-    @Override
-    public void onDrmSessionManagerError(Exception e) {
```
juankysoriano: According to the documentation:

```
    /**
     * Called when a drm error occurs.
     * <p>
     * This method being called does not indicate that playback has failed, or that it will fail.
     * The player may be able to recover from the error and continue. Hence applications should
     * <em>not</em> implement this method to display a user visible error or initiate an application
     * level retry ({@link Player.EventListener#onPlayerError} is the appropriate place to implement
     * such behavior). This method is called to provide the application with an opportunity to log
     * the error if it wishes to do so.
     *
     * @param e The corresponding exception.
     */
```

So we shouldn't be delegating to `PlayerError` for client handling and instead of that simply logging

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/MediaSourceOnErrorForwarder.java
```diff
@@ -1,25 +0,0 @@
...
-class MediaSourceOnErrorForwarder implements ExtractorMediaSource.EventListener {
-
-    private final NoPlayer.ErrorListener errorListener;
-
-    MediaSourceOnErrorForwarder(NoPlayer.ErrorListener errorListener) {
-        this.errorListener = errorListener;
-    }
-
-    @Override
-    public void onLoadError(IOException error) {
```
juankysoriano: According to the documentation:

```
    /**
     * Called when a drm error occurs.
     * <p>
     * This method being called does not indicate that playback has failed, or that it will fail.
     * The player may be able to recover from the error and continue. Hence applications should
     * <em>not</em> implement this method to display a user visible error or initiate an application
     * level retry ({@link Player.EventListener#onPlayerError} is the appropriate place to implement
     * such behavior). This method is called to provide the application with an opportunity to log
     * the error if it wishes to do so.
     *
     * @param e The corresponding exception.
     */
```

So we shouldn't be delegating to `PlayerError` for client handling and instead of that simply logging

core/src/test/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerErrorMapperTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+import static com.google.android.exoplayer2.upstream.HttpDataSource.InvalidResponseCodeException;
+import static org.fest.assertions.api.Assertions.assertThat;
+
+@RunWith(Enclosed.class)
+public class ExoPlayerErrorMapperTest {
+
+    public static class GivenSourceException {
+        @Test
+        public void InvalidResponseCodeException_mapsTo_InvalidResponseCode() {
```
Mecharyry: Uppercase 😬

juankysoriano: It is upper case because it is the name of the class `InvalidResponseCode`

ouchadam:
:+1: sounds promising

PR #137, opened 2018-03-29
Support Webvtt subtitles class colors

danybony:
## Problem

Colors defined in Webvtt cue Class tags were ignored.
For example in the screenshots below the cue text was
```
At the <c.red.caps>left</c> we can see...
```
Example asset: http://dash.edgesuite.net/akamai/test/caption_test/ElephantsDream/elephants_dream_480p_heaac5_1.mpd

## Solution

The core of the change is done in the commit https://github.com/novoda/no-player/commit/72095c6d329c10c57bfb98ff89bbf81829897f04.
Then, to make it available, we had to import a bunch of classes related to `WebvttCueParser`, which were still package protected in Exoplayer, and to make our `ExoPlayerCreator` use a custom SubtitleDecoderFactory. This factory is exactly the same as the default one, with the only change in the import for our modified `WebvttCueParser`. 

I suppressed static analysis for the imported classes and solved a Lint warning related to outdated Mockito dependency (updated it to `2.16.0`, release notes at https://github.com/mockito/mockito/blob/release/2.x/doc/release-notes/official.md)

### Test(s) added 

No

### Screenshots

| Before | After |
| ------ | ----- |
| pasted image at 2018_03_29 11_46 am |  pasted image at 2018_03_29 11_43 am |

### Paired with 

Me and @ouchadam investigated this together

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -22,7 +22,7 @@ dependencies {
     implementation 'com.google.android.exoplayer:exoplayer:2.7.1'
 
     testImplementation 'junit:junit:4.12'
-    testImplementation 'org.mockito:mockito-core:2.15.0'
+    testImplementation 'org.mockito:mockito-core:2.16.0'
     testImplementation 'org.easytesting:fest-assert-core:2.0M10'
 }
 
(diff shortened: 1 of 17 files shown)


ouchadam:
How do you feel about moving the copied `exoplayer` classes to `exoplayer/external`?

Issue #111, opened 2017-11-14
Add api to change volume/mute


ouchadam:
#### Problem

We do not currently expose setting the player volume.

#### Impact

- Need investigate behaviour of mediaplayer vs exoplayer

PR #139, opened 2018-04-03
NO-111: Add API to change volume

JozefCeluch:
## Problem

There was no API to control video player volume (issue #111)

## Solution

I added the `getVolume`/`setVolume()` methods to the `NoPlayer` interface and then implemented the chain all the way down to each `Facade`.

In case of `ExoPlayer` this was just the case of exposing the volume controls on the `NoPlayer` API. 

The implementation is slightly different for `MediaPlayer` since it doesn't provide `getVolume()` method. In order to achieve that I added the volume state that defaults to 1 into the `Facade` (which is exactly what `ExoPlayer` does internally anyway).

I added the volume toggle into the demo app.

### Test(s) added 

all new logic should be tested

### Screenshots

| Before | Volume off | Volume on |
| ------ | ----- | ----- |
| image | image | image |

### Paired with 

Nobody


core/src/main/java/com/novoda/noplayer/NoPlayer.java
```diff
@@ -205,6 +206,22 @@
      */
     void setRepeating(boolean repeating) throws IllegalStateException;
 
+    /**
+     * Set the audio volume, with 0 being silence and 1 being unity gain.
+     *
+     * @param volume The audio volume.
+     * @throws IllegalStateException - if called before {@link NoPlayer#loadVideo(Uri, ContentType)}.
+     */
+    void setVolume(@FloatRange(from = 0.0f, to = 1.0f) float volume);
```
ouchadam: nice use of the Range! 💯

JozefCeluch:
@ouchadam good question, the default behavior we get right now from ExoPlayer is that the volume gets reset to 1 when we load the video (since we create a new ExoPlayer instance). I would just go with that behavior for now in `MediaPlayerFacade` too. It reduces the amount of state we need to keep

ouchadam:
@JozefCeluch sounds good 👍 We can always keep track of the state internally if we decide we do want that experience

PR #142, opened 2018-04-16
Only emit beat when the player is playing

Mecharyry:
## Problem
We've seen in a client application that it is possible for an `onBeat` to be emitted when the player is no longer playing.

## Solution
As the `Heart` uses a `Handler` we should check that the player is indeed playing in the `Heart.Heartbeat` before emitting an `onBeat`. 

### Test(s) added 
Yes, tested the `Heart` class.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/Heart.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/Heart.java
@@ -78,7 +78,9 @@ public class Heart {
 
         @Override
         public void run() {
-            callback.onBeat(player);
+            if (player.isPlaying()) {
+                callback.onBeat(player);
+            }
         }
     }
 }
--- a/core/src/test/java/com/novoda/noplayer/internal/HeartTest.java
+++ b/core/src/test/java/com/novoda/noplayer/internal/HeartTest.java
@@ -1,0 +1,102 @@
+package com.novoda.noplayer.internal;
+
+import android.os.Handler;
+
+import com.novoda.noplayer.NoPlayer;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.BDDMockito.then;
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/novoda/noplayer/internal/Heart.java
```diff
@@ -78,7 +78,9 @@ public Heartbeat(NoPlayer.HeartbeatCallback callback, NoPlayer player) {
 
         @Override
         public void run() {
-            callback.onBeat(player);
+            if (player.isPlaying()) {
```
rock3r: Why not starting/stopping the heartbeat when the playback starts/end instead? Seems more straightforward

Mecharyry: We do that already. The problem is that this handler might execute between player state changes 😢 it's kind of an edge case but we have seen it.

ouchadam: do you think we need to log/call a listener to let clients know this heartbeat has been skipped?

Mecharyry: I'm not sure tbh, I think emitting when it isn't playing is an error, it shouldn't happen. We would just be telling clients about behaviour we would never expect them to deal with. What do you think?

ouchadam: I agree, just a little paranoid that we might have changed behaviour (by fixing bad state).

I'd be happy with release notes that explain the scenario

core/src/test/java/com/novoda/noplayer/internal/HeartTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.mockito.stubbing.Answer;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.BDDMockito.then;
+import static org.mockito.BDDMockito.will;
+import static org.mockito.Mockito.mock;
+
+public class HeartTest {
```
ouchadam: :100:

Mecharyry:
Happy to merge @rock3r ?

rock3r:
Yep works for me, if it's ok for Adam who knows the ins and outs :)

PR #143, opened 2018-04-18
Add missing throws declaration from volume api

Mecharyry:
## Problem
I noticed when implementing some changes on a client application that we are missing the `throws` declaration on the volume api.

## Solution
Add the `throws` declaration.

### Paired with 
Nobody.

### DIFF
--- a/core/src/test/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacadeTest.java
+++ b/core/src/test/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacadeTest.java
@@ -45,13 +45,15 @@ import static org.fest.assertions.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.BDDMockito.given;
-import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 @RunWith(Enclosed.class)
 public class ExoPlayerFacadeTest {
 
     private static final boolean SELECTED = true;
-    private static final PlayerVideoTrack PLAYER_VIDEO_TRACK = PlayerVideoTrackFixture.aPlayerVideoTrack().build();
 
     private static final long TWO_MINUTES_IN_MILLIS = 120000;
     private static final long TEN_MINUTES_IN_MILLIS = 600000;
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacadeTest.java
```diff
@@ -206,6 +208,20 @@ public void selectSubtitleTrack_thenThrowsIllegalStateException() {
...
+
+        @Test
+        public void whenSetVolume_thenThrowsIllegalStateException() {
+            thrown.expect(ExceptionMatcher.matches("Video must be loaded before trying to interact with the player", IllegalStateException.class));
+
+            facade.setVolume(ANY_VOLUME);
+        }
+
+        @Test
+        public void whenGetVolume_thenthenThrowsIllegalStateException() {
```
takecare: extra `then`

core/src/test/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacadeTest.java
```diff
@@ -206,6 +208,20 @@ public void selectSubtitleTrack_thenThrowsIllegalStateException() {
 
             facade.selectSubtitleTrack(subtitleTrack);
         }
+
+        @Test
+        public void whenSetVolume_thenThrowsIllegalStateException() {
```
takecare: `whenSetVolumeWithoutLoadingVideoFirst_thenThrowsIllegalStateException`

core/src/test/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacadeTest.java
Mecharyry: I'd avoid the `withoutLoadingVideoFirst` because the enclosing class is named `GivenVideoNotLoaded`

core/src/test/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacadeTest.java
Mecharyry: Same as above, the enclosing class is `GivenVideoNotLoaded`

core/src/test/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacadeTest.java
Mecharyry: 26a688c

PR #144, opened 2018-04-20
Track selectors create at loadVideo

MechaRyRy:
Mainly opening to prompt a discussion with @ouchadam and @JozefCeluch, we were talking about this the other day.

## Problem
We would like the ability to change the `TrackSelector`s for `ExoPlayer` at the time we load the video. Creating the `TrackSelector`s at this point will allow us to specify rendition quality check times for each individual video we load, rather than just for a player.

## Solution
This PR is just refactoring, we will build on this in a follow-up PR: I've extracted all the `TrackSelector`s to a `CompositeTrackSelector`. The `CompositeTrackSelector` is created at the time the video is loaded and passed to the `ExoPlayerCreator`. 

## Next steps
Add more params to the player builder that will influence the track selectors.

### Paired with 
Nobody

### DIFF
--- a/core/src/test/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacadeTest.java
+++ b/core/src/test/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacadeTest.java
@@ -13,9 +13,6 @@ import com.google.android.exoplayer2.source.MediaSource;
 import com.novoda.noplayer.ContentType;
 import com.novoda.noplayer.internal.exoplayer.drm.DrmSessionCreator;
 import com.novoda.noplayer.internal.exoplayer.forwarder.ExoPlayerForwarder;
-import com.novoda.noplayer.internal.exoplayer.mediasource.ExoPlayerAudioTrackSelector;
-import com.novoda.noplayer.internal.exoplayer.mediasource.ExoPlayerSubtitleTrackSelector;
-import com.novoda.noplayer.internal.exoplayer.mediasource.ExoPlayerVideoTrackSelector;
 import com.novoda.noplayer.internal.exoplayer.mediasource.MediaSourceFactory;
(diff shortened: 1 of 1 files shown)


ouchadam:
I really dig it _buttttt_ do we now lose the selected track state when  loadVideo -> stop -> loadVideo (did we lose it before anyways?)

zegnus:
@Mecharyry looks like this PR is just refactoring, neither adds or remove any functionality. How does this PR solve the issue?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -114,7 +107,8 @@ void loadVideo(SurfaceHolder surfaceHolder,
                    ExoPlayerForwarder forwarder,
                    MediaCodecSelector mediaCodecSelector) {
         this.contentType = contentType;
-        exoPlayer = exoPlayerCreator.create(drmSessionCreator, forwarder.drmSessionEventListener(), mediaCodecSelector);
+        trackSelector = trackSelectorCreator.create();
```
ouchadam: @zegnus this line is where we now creation the track selector, rather than having all the track selectors precreated

JozefCeluch:
👍 nice!

Mecharyry:
> Add more params to the player builder that will influence the track selectors.

@zegnus in the PR description, this is just the beginning. I can create a feature branch for this work if you would all like? @ouchadam @JozefCeluch ?

Mecharyry:
> I really dig it buttttt do we now lose the selected track state when loadVideo -> stop -> loadVideo (did we lose it before anyways?)

Good question, I think we lost it anyway. I'll add some additional buttons to the demo and check this out as soon as I can.

ouchadam:
:+1: for a feature branch

PR #150, opened 2018-05-08
NO-149/SurfaceRequester

Mecharyry:
## Problem
The current implementation of `NoPlayer` relies upon a `SurfaceHolderRequester` to pass a `SurfaceHolder` to the underlying player. Clients may want to use `SurfaceView`, `TextureView` etc and the underlying player, `ExoPlayer` specifically only cares about the `Surface`.

## Solution
Switch out the `SurfaceHolderRequester` with a `SurfaceRequester` passing a `Surface` to the underlying player. 

### Paired with 
Nobody.


core/src/main/java/com/novoda/noplayer/FramesPerSecondCalculator.java
```diff
@@ -0,0 +1,35 @@
+package com.novoda.noplayer;
+
+import java.util.Deque;
+import java.util.LinkedList;
+import java.util.concurrent.TimeUnit;
+
```
ouchadam: should this class be in this PR?

mr-archano: Yeah, I would consider to remove it @Mecharyry, just to avoid to ship it in what is effectively a non-related change set.

core/src/main/java/com/novoda/noplayer/SurfaceRequester.java
```diff
@@ -0,0 +1,17 @@
+package com.novoda.noplayer;
+
+import android.view.Surface;
+
+public interface SurfaceRequester {
+
+    void requestSurfaceTexture(Callback callback);
```
mr-archano: No need to use texture, right?

core/src/main/res/layout/noplayer_view.xml
```diff
@@ -5,7 +5,7 @@
   android:layout_height="wrap_content">
 
   <SurfaceView
-    android:id="@+id/surface_view"
+    android:id="@+id/texture_view"
```
mr-archano: shouldn't we keep this as it was?

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -75,17 +75,20 @@ public void onClick(View v) {
     };
 
     private final View.OnClickListener showAudioSelectionDialog = new View.OnClickListener() {
-
         @Override
         public void onClick(View v) {
-            dialogCreator.showAudioSelectionDialog();
+            if (player == null || !player.isPlaying() || player.getAudioTracks().size() == 0) {
+                Toast.makeText(MainActivity.this, "no audio tracks available!", Toast.LENGTH_LONG).show();
```
mr-archano: Why this change?

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -75,17 +75,20 @@ public void onClick(View v) {
...
+                dialogCreator.showAudioSelectionDialog();
+            }
         }
     };
 
     private final View.OnClickListener showSubtitleSelectionDialog = new View.OnClickListener() {
         @Override
         public void onClick(View v) {
-            if (player.getSubtitleTracks().isEmpty()) {
+            if (player == null || !player.isPlaying() || player.getSubtitleTracks().isEmpty()) {
```
mr-archano: Same question here (I know that this is a demo, I am just curious)

demo/src/main/java/com/novoda/demo/MainActivity.java
Mecharyry: The buttons to trigger the dialog are always present, if you tap them before the player is playing then it will throw an `IllegalStateException`.

demo/src/main/java/com/novoda/demo/MainActivity.java
Mecharyry: Same as above 😄

core/src/main/java/com/novoda/noplayer/FramesPerSecondCalculator.java
Mecharyry: 728d499

core/src/main/java/com/novoda/noplayer/SurfaceRequester.java
Mecharyry: 20b70e5

core/src/main/res/layout/noplayer_view.xml
Mecharyry: 899678b

core/src/main/java/com/novoda/noplayer/NoPlayerView.java
```diff
@@ -51,7 +51,7 @@ public View getContainerView() {
     }
 
     @Override
-    public SurfaceHolderRequester getSurfaceHolderRequester() {
+    public SurfaceRequester getSurfaceTextureRequester() {
```
ouchadam: Do we need to leak `texture` here? SurfaceRequester should be enough?

Mecharyry: 6312a81

PR #152, opened 2018-05-10
Support snapshot releases from `develop`

mr-archano:
## Problem

When we work on a new feature or a bug fix we would like to be able to use a new snapshot of the library without releasing to JCenter. 

## Solution

We have decided to use a new Bintray repo (https://bintray.com/novoda/snapshots) where to push such builds. This PR adds the necessary Gradle setup that allows us to define whether a certain build on the CI needs to land to the snapshots repo or not.

A [new CI job](https://ci.novoda.com/job/no-player-snapshot/configure) has been configured:
- it should be triggered only when `develop` changes (ie: after a feature branch is merged)
- it should always push to https://bintray.com/novoda/snapshots

### Test(s) added 

It's all Gradle stuff, so no automated test. I had to check manually whether the inputs were correctly handled in the build script.

### Screenshots

No UI changes, therefore no screenshots.

### Paired with 

Nobody

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -1,6 +1,31 @@
 apply plugin: 'com.android.library'
 apply plugin: 'bintray-release'
 apply plugin: 'jacoco'
+apply plugin: 'com.novoda.build-properties'
+
+buildProperties {
+    cli {
+        using(project)
+    }
+    bintray {
+        def bintrayCredentials = {
+            boolean isDryRun = cli['dryRun'].or(true).boolean
+            return isDryRun ?
+                    ['bintrayRepo': 'n/a', 'bintrayUser': 'n/a', 'bintrayKey': 'n/a'] :
+                    new File("${System.getenv('BINTRAY_PROPERTIES')}")
+        }
+        using(cli).or(bintrayCredentials())
+        description = '''This should contain the following properties:
+                       | - bintrayRepo: name of the repo of the organisation to deploy the artifacts to
+                       | - bintrayUser: name of the account used to deploy the artifacts
+                       | - bintrayKey: API key of the account used to deploy the artifacts
+        '''.stripMargin()
+    }
+    publish {
+        using(['version': version])
+                .or(buildProperties.bintray)
+    }
+}
 
 android {
     compileSdkVersion 27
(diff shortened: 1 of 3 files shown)


core/build.gradle
```diff
@@ -1,6 +1,35 @@
...
+        using(project)
+    }
+    bintray {
+        def bintrayCredentials = {
+            boolean isDryRun = cli['dryRun'].or(true).boolean
+            return isDryRun ?
+                    ['bintrayRepo': 'n/a', 'bintrayUser': 'n/a', 'bintrayKey': 'n/a'] :
+                    new File("${System.getenv('BINTRAY_PROPERTIES')}")
+        }
+        using(cli).or(bintrayCredentials())
```
ouchadam: this is really cool :sunglasses:

core/build.gradle
```diff
@@ -1,6 +1,35 @@
...
+        description = '''This should contain the following properties:
+                       | - bintrayRepo: name of the repo of the organisation to deploy the artifacts to
+                       | - bintrayUser: name of the account used to deploy the artifacts
+                       | - bintrayKey: API key of the account used to deploy the artifacts
+        '''.stripMargin()
+    }
+    publish {
+        def generateVersion = {
+            boolean isSnapshot = cli['bintraySnapshot'].or(false).boolean
+            return isSnapshot ? System.getenv('GIT_COMMIT')?:'SNAPSHOT' : version
```
ouchadam: git commit or jenkins build number? 

I'm leaning towards build number as it's easier to communicate

ouchadam:
small question around the variable to build the version out of, otherwise looking great! :100:

core/build.gradle
mr-archano: yeah, I had no idea tbh :D
we can pick other values [too](https://wiki.jenkins.io/display/JENKINS/Building+a+software+project)), but I am happy to go for `BUILD_NUMBER` then

Mecharyry:
screen shot 2018-05-11 at 09 08 35
Is it worth us adding a badge to the repo, pointing to snapshots?

ouchadam:
good point about the badges.
It might be easier to have both master and snapshot badges

Issue #149, opened 2018-05-08
Add support for TextureView


mr-archano:
#### Problem

At the moment the integration of `ExoPlayer`/`MediaPlayer` is relying on the client to use a `SurfaceView` as host of the video buffer. This entails a rendering performance problem when the video is hosted in scrollable containers (eg: a `RecyclerView`), a known issue as described in the [official documentation from `ExoPlayer`](https://google.github.io/ExoPlayer/faqs.html#how-do-i-get-smooth-animationscrolling-of-video):
> `SurfaceView` rendering wasn’t properly synchronized with view animations until Android N. On earlier releases this could result in unwanted effects when a `SurfaceView` was placed into scrolling container, or when it was subjected to animation. Such effects included the `SurfaceView`’s contents appearing to lag slightly behind where it should be displayed, and the view turning black when subjected to animation.

#### Potential Solution

The `no-player` API is providing facades for both `ExoPlayer` and `MediaPlayer`, but one of the callbacks that clients have to provide makes use of [`android.view.SurfaceHolder`](https://developer.android.com/reference/android/view/SurfaceHolder), a component that is tightly coupled to the `SurfaceView` lifecycle.
Both media frameworks provide APIs that accept a [`Surface`](https://developer.android.com/reference/android/view/Surface) in lieu of `SurfaceHolder`, altho the support in `MediaPlayer` for those looks to be API 14+ (still below the minsdk of the library afaict).

It should be relatively easy to change the implementation of [`SurfaceHolderRequester.Callback`](https://github.com/novoda/no-player/blob/3307272f21a81919dc230aead3791a93809e330f/core/src/main/java/com/novoda/noplayer/SurfaceHolderRequester.java#L11-L15) to deprecate the use of `SurfaceHolder` and introduce a new callback accepting a plain `Surface` instead.

#### Impact

Promoting the use of `Surface` instead of `SurfaceHolder` will allow the consumers to choose `SurfaceView` or `TextureView` on their side without any repercussion on the player itself.
As suggested above we could try to go for duplication + deprecation as first step, but afaict clients will be affected by the change anyway, and adjusting to the new API is pretty trivial (using `SurfaceHolder.getSurface()` to retrieve the surface to pass to the callback.

Mecharyry:
I'll start taking a look at this and track against https://github.com/novoda/no-player/tree/NO-149/surface_feature and https://github.com/novoda/no-player/tree/NO-149/surface

Mecharyry:
[...]

PR #154, opened 2018-05-11
NO-149: Remove hard dependency on SurfaceHolder

mr-archano:
## Problem

Given #150 has been reviewed, we would like to create a snapshot build that allows the use of `TextureView`.

## Solution

Just open a PR against `develop`. The code has been already reviewed in #150.
This will address #149.


mr-archano:
Good point @ouchadam, how do we want to track this? We could even update the changelog as part of this PR... 🤔

Mecharyry:
What's the plan for the above?

ouchadam:
I'm easy, was thinking the release notes, as long as we have it somewhere.

PR #157, opened 2018-05-25
Track selection fix

Mecharyry:
## Problem
We created a wrapper around the `TrackSelector` but we did something weird where we implemented the interface **and** used a `DefaultTrackSelector` the default was used for everything but we passed the wrapper to the `ExoPlayer`. This meant that the track selection wasn't working 😬 

## Solution
`CompositeTrackSelector` is now a pure wrapper, it doesn't extend any interfaces. We expose the `TrackSelector` that the `ExoPlayer` needs through a getter instead. 

### Test(s) added 
Just updated some. I've manually tested this solution on a client app and the track selection is working again.

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelector.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelector.java
@@ -1,12 +1,8 @@
 package com.novoda.noplayer.internal.exoplayer;
 
-import com.google.android.exoplayer2.ExoPlaybackException;
-import com.google.android.exoplayer2.RendererCapabilities;
 import com.google.android.exoplayer2.SimpleExoPlayer;
-import com.google.android.exoplayer2.source.TrackGroupArray;
 import com.google.android.exoplayer2.trackselection.DefaultTrackSelector;
 import com.google.android.exoplayer2.trackselection.TrackSelector;
-import com.google.android.exoplayer2.trackselection.TrackSelectorResult;
 import com.novoda.noplayer.ContentType;
 import com.novoda.noplayer.internal.exoplayer.mediasource.ExoPlayerAudioTrackSelector;
 import com.novoda.noplayer.internal.exoplayer.mediasource.ExoPlayerSubtitleTrackSelector;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelector.java
```diff
@@ -19,7 +15,7 @@
 
 import java.util.List;
 
-class CompositeTrackSelector extends TrackSelector {
```
Mecharyry: We could probably extend the `DefaultTrackSelector` but it is also passed to the other track selectors 😬 I'll look at updating this again on dev, this is a quick fix to ensure that master works.

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -16,7 +16,7 @@
 
 public class MainActivity extends Activity {
 
-    private static final String URI_VIDEO_WIDEVINE_EXAMPLE_MODULAR_MPD = "https://storage.googleapis.com/wvmedia/cenc/h264/tears/tears.mpd";
```
Mecharyry: This asset wasn't very useful, it didn't have any audio tracks 😬 we need some custom content or at least an asset selection to make testing easier in the demo.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelector.java
mr-archano: it feels to me we are missing an interface somewhere

demo/src/main/java/com/novoda/demo/MainActivity.java
ouchadam: 👍 we should definitely create our own samples

core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelector.java
```diff
@@ -77,14 +77,4 @@ boolean selectTextTrack(PlayerSubtitleTrack subtitleTrack, RendererTypeRequester
     boolean clearSubtitleTrack(RendererTypeRequester rendererTypeRequester) {
         return subtitleTrackSelector.clearSubtitleTrack(rendererTypeRequester);
     }
-
-    @Override
-    public TrackSelectorResult selectTracks(RendererCapabilities[] rendererCapabilities, TrackGroupArray trackGroups) throws ExoPlaybackException {
```
ouchadam: I don't fully understand why delegate wasn't working?  🤔

Mecharyry:
Wait, I might have found the issue

PR #160, opened 2018-06-01
Support for TextureView

JozefCeluch:
## Problem

NoPlayer did not support `TextureView` because `PlayerView` was quite tightly coupled with `SurfaceHolder` which is not available in `TextureView`.

This is an alternative approach to the one that was tried in #150 which attempted to replace `SurfaceHolder` with `Surface`. Unfortunately that had some unexpected side-effects because ExoPlayer internally observes the changes in `SurfaceHolder` which were not accounted for.

## Solution

This approach is slightly different in that it explicitly adds support for both `TextureView` and `SurfaceView` to `PlayerView`. The client needs to provide an instance of `SurfaceContainer` that can contain either one of the views.

Internally, the ExoPlayer implementation uses the views directly and the MediaPlayer implementation observes and uses the `Surface` in a very much the same way as in #150 (with the difference that this is now an internal implementation detail that the clients don't need to know about). There might be some problems (or different behavior) with this approach in MediaPlayer implementation because internally it behaves differently when it contains `SurfaceHolder` and when it only contains `Surface`. For example setting screen on while playing works only when `SurfaceHolder` is provided ([source](https://github.com/aosp-mirror/platform_frameworks_base/blob/186be90eb6dbbf1fe1ad28c8faca47e9b80997d8/media/java/android/media/MediaPlayer.java#L783))

### Test(s) added 

updated where necessary

### Screenshots

no UI changes

### Paired with 

nobody


core/src/main/java/com/novoda/noplayer/PlayerView.java
```diff
@@ -19,4 +21,57 @@
     void hideSubtitles();
 
     void setSubtitleCue(TextCues textCues);
+
+    class SurfaceContainer {
```
JozefCeluch: not sure about the name of this thing, also if it should be inside the interface

core/src/main/java/com/novoda/noplayer/PlayerView.java
```diff
@@ -19,4 +21,57 @@
...
+            return new SurfaceContainer(surfaceView, null, surfaceHolder);
+        }
+
+        public static SurfaceContainer create(TextureView textureView) {
+            PlayerViewSurfaceHolder surfaceHolder = new PlayerViewSurfaceHolder();
+            textureView.setSurfaceTextureListener(surfaceHolder);
+            return new SurfaceContainer(null, textureView, new PlayerViewSurfaceHolder());
+        }
+
+        SurfaceContainer(@Nullable SurfaceView surfaceView, @Nullable TextureView textureView, PlayerViewSurfaceHolder surfaceHolder) {
```
rock3r: Shouldn't this be private?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -139,8 +138,14 @@ private void setMovieAudioAttributes(SimpleExoPlayer exoPlayer) {
...
 
-    private void attachToSurface(SurfaceHolder surfaceHolder) {
-        exoPlayer.setVideoSurfaceHolder(surfaceHolder);
+    private void attachToSurface(PlayerView.SurfaceContainer surfaceContainer) {
+        if (surfaceContainer.containsSurfaceView()) {
+            exoPlayer.setVideoSurfaceView(surfaceContainer.getSurfaceView());
+        } else if (surfaceContainer.containsTextureView()) {
+            exoPlayer.setVideoTextureView(surfaceContainer.getTextureView());
+        } else {
+            throw new IllegalStateException("Surface container does not contain any of the expected views");
```
rock3r: Looks more like an invalid argument than an invalid state to me

core/src/main/java/com/novoda/noplayer/PlayerView.java
JozefCeluch: I tried... adding _Player_ at the beginning makes it not confusing at all, right right

core/src/main/java/com/novoda/noplayer/PlayerView.java
JozefCeluch: then the PMD and checkstyle complain that the class should be `final` and then I cannot mock it in tests

Mecharyry: You can mock final classes in tests, there is a hack to do it, I thought we had it in `no-player` 🤔

Nope, turns out we have it in `download-manager`
 https://github.com/novoda/download-manager/blob/epic_download_manager_v2/library/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker

https://github.com/mockito/mockito/wiki/What's-new-in-Mockito-2#mock-the-unmockable-opt-in-mocking-of-final-classesmethods

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -139,8 +138,14 @@ private void setMovieAudioAttributes(SimpleExoPlayer exoPlayer) {
         }
     }
 
-    private void attachToSurface(SurfaceHolder surfaceHolder) {
-        exoPlayer.setVideoSurfaceHolder(surfaceHolder);
+    private void attachToSurface(PlayerSurfaceHolder playerSurfaceHolder) {
+        if (playerSurfaceHolder.containsSurfaceView()) {
```
mr-archano: I would prefer to avoid this check, maybe better to have something like `playerSurfaceHolder.install(exoPlayer)`

JozefCeluch: ooooh yeah, nice! I'll do that

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -139,8 +138,14 @@ private void setMovieAudioAttributes(SimpleExoPlayer exoPlayer) {
...
+            exoPlayer.setVideoSurfaceView(playerSurfaceHolder.getSurfaceView());
+        } else if (playerSurfaceHolder.containsTextureView()) {
+            exoPlayer.setVideoTextureView(playerSurfaceHolder.getTextureView());
```
ouchadam: do we know what the difference was between `set*View` and `setSurface` that was causing the glitchy behaviour in the previous solution?

mr-archano: afaict internally the exoplayer holds a field for either a surface holder or a surface, I reckon that there's some different code path where a holder is used in lieu of a surface

core/src/main/java/com/novoda/noplayer/NoPlayerView.java
```diff
@@ -51,8 +47,8 @@ public View getContainerView() {
     }
 
     @Override
-    public SurfaceHolderRequester getSurfaceHolderRequester() {
-        return surfaceHolderProvider;
+    public PlayerSurfaceHolder getPlayerSurfaceHolder() {
+        return PlayerSurfaceHolder.create(surfaceView);
```
ouchadam: Small nitpick but `getters` which do more than `get` should probably have a different name :smile_cat:   

Do we need to create a new `PlayerSurfaceHolder` each time the `get` is called or could we create it as part of `onFinishInflate`?

JozefCeluch: Well spotted! It was the thing I _was planning to remember to move later_ 😄

PR #161, opened 2018-06-05
Update ExoPlayer to latest v2.8.1

zegnus:
## Problem

The latest ExoPlayer v2.8.1 has lots of breaking changes, we want to use this latest version so that we are on par with the latest improvements and bug fixes

- Deprecating `DefaultDrmSessionManager.EventListener`, should now use `DefaultDrmSessionEventListener` instead.

- `RendererFactories` now exposes the `DrmSessionManager` through the `createRenderers` method so there is no need to pass it through the constructor. This means we now pass the `DrmSessionManager` through to the `ExoPlayerFactory.newSimpleInstance` instead.

- `ExoPlayer.setVideoDebugListener` has been deprecated, use `addAnalyticsListener` instead.

- `DefaultBandwidthMeter` should make use of `Options.maxInitialBitrate` wherever it is used.

- `AdaptiveMediaSourceEventListener` has been deprecated, use `MediaSourceEventListener` instead.

- Extracted `ForwarderInformation.Methods` and `ForwarderInformation.Parameters` to hold the keys
used by the info forwarders.

- `TrackAudioSelector` can make use of the `DefaultTrackSelector` rather than the `MappingTrackSelector`, dropping the need for the `TrackSelectionFactory`. `TrackSelectors` now set parameters building upon what is present rather than resetting and building from scratch. 

- Delay creation of the `DataSourceFactory` until we actually need to create a media source, this allows us to use `Options` rather than passing additional parameters through the higher level
creators.

## Solution

Update the library and solve all the breaking changes

### Test(s) added 

All updated

### Screenshots

All works as before, no UI changes

### Paired with 

@Mecharyry


core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelectorCreator.java
```diff
@@ -14,16 +13,13 @@
...
 class CompositeTrackSelectorCreator {
 
-    private final DefaultBandwidthMeter bandwidthMeter;
-
-    CompositeTrackSelectorCreator(DefaultBandwidthMeter bandwidthMeter) {
-        this.bandwidthMeter = bandwidthMeter;
-    }
-
     CompositeTrackSelector create(Options options) {
+        DefaultBandwidthMeter bandwidthMeter = new DefaultBandwidthMeter.Builder()
```
Mecharyry: Should be good to have this separate from the other one from what I have seen. 🏈

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/MediaSourceFactory.java
```diff
@@ -1,60 +1,74 @@
...
     }
 
-    public MediaSource create(ContentType contentType,
+    public MediaSource create(Options options,
                               Uri uri,
-                              ExtractorMediaSource.EventListener eventListener,
-                              AdaptiveMediaSourceEventListener mediaSourceEventListener) {
-        switch (contentType) {
+                              MediaSourceEventListener mediaSourceEventListener) {
+        DefaultBandwidthMeter bandwidthMeter = new DefaultBandwidthMeter.Builder()
```
Mecharyry: 🏈 The second bandwidth meter. It seems that it doesn't need to be the same instance.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/MediaSourceFactory.java
```diff
@@ -1,60 +1,74 @@
...
+                .setInitialBitrateEstimate(options.maxInitialBitrate())
+                .build();
+        DefaultDataSourceFactory defaultDataSourceFactory = new DefaultDataSourceFactory(context, "user-agent", bandwidthMeter);
```
Mecharyry: Delayed creation of this and the `BandwidthMeter` so that we can make use of the `Options` to create the `BandwidthMeter` rather than passing a param as part of the `Factories` and `Creators`.

ouchadam:
any chance of an overview in the PR description? :angel:

Mecharyry:
@ouchadam I've updated the PR description to show the main changes that we have done 😄 let me know if you need anything else.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelectorCreator.java
ouchadam: Whilst we have more control over rendition selecting across sources it also means we have multiple places to change. 

How  do you feel about introduce a holder for audio/video/text bandwidth meters? and potentially be default we could reuse the same one

Mecharyry: I opted to pass through in the two places it is used, they both get called in the same place on the Facade 😄 also helpful to ensure that it is the same instance, it might have some affect that I haven't accounted for 😄

ouchadam: :100: nice! I'm not sure about two independent instances yet, it might be safer to test first but with a single place to pass them from I'm happy

PR #162, opened 2018-06-11
Provide both Surface and SurfaceHolder in MediaPlayer

JozefCeluch:
## Problem

While adding a support for `TextureView` I (possibly) changed the behavior of `MediaPlayer` with the `SurfaceView` because instead of providing `SurfaceHolder` to MediaPlayer I started providing `Surface`. I don't know if that changes any behavior, so it's probably for the best to keep the changes to minimum.

## Solution

Add support to provide both `Surface` and `SurfaceHolder` to MediaPlayer by introducing a concept of `Either`. That way, the `Surface` is only provided if no-player is used with `TextureView` which is completely separate from the previous `SurfaceView` behavior that uses `SurfaceHolder`.

### Test(s) added 

updated and added where necessary

### Screenshots

no UI changes

### Paired with 

nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
@@ -6,12 +6,13 @@ import android.media.MediaPlayer;
 import android.net.Uri;
 import android.support.annotation.Nullable;
 import android.view.Surface;
-
+import android.view.SurfaceHolder;
 import com.novoda.noplayer.internal.mediaplayer.PlaybackStateChecker.PlaybackState;
 import com.novoda.noplayer.internal.mediaplayer.forwarder.MediaPlayerForwarder;
 import com.novoda.noplayer.internal.utils.NoPlayerLog;
 import com.novoda.noplayer.internal.utils.Optional;
 import com.novoda.noplayer.model.AudioTracks;
+import com.novoda.noplayer.model.Either;
 import com.novoda.noplayer.model.PlayerAudioTrack;
 import com.novoda.noplayer.model.PlayerSubtitleTrack;
 import com.novoda.noplayer.model.PlayerVideoTrack;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -173,13 +174,29 @@ void release() {
...
+                mediaPlayer.setSurface(value);
+            }
+        };
+        Either.Consumer<SurfaceHolder> setDisplay = new Either.Consumer<SurfaceHolder>() {
+            @Override
+            public void accept(SurfaceHolder value) {
+                mediaPlayer.setDisplay(value);
+            }
+        };
+        surface.apply(setSurface, setDisplay);
```
ouchadam: really digging this :+1:

core/src/main/java/com/novoda/noplayer/model/Either.java
```diff
@@ -0,0 +1,50 @@
+package com.novoda.noplayer.model;
```
ouchadam: Are we exposing this to clients? If not the project is using the convention of `internal` to describe public classes which do not provide a guaranteed api.

JozefCeluch: Not directly, but it's used in `SurfaceRequester` which is kinda exposed in the public API through `PlayerSurfaceHolder`(the clients need to provide an instance of that when they implement a custom implementation of `PlayerView`).

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -67,7 +68,7 @@ static AndroidMediaPlayerFacade newInstance(Context context, MediaPlayerForwarde
         this.mediaPlayerCreator = mediaPlayerCreator;
     }
 
-    void prepareVideo(Uri videoUri, Surface surface) {
+    public void prepareVideo(Uri videoUri, Either<Surface, SurfaceHolder> surface) {
```
Mecharyry: Think this `public` was done in error, the class is still package.

PR #169, opened 2018-07-12
Fix video changed listener

joetimmins:
## Problem
The method on `exoplayer` to attach a `VideoRendererEventListener` has been deprecated. That means we weren't attaching our `VideoSizeChangedListener` properly.

## Solution
Change the forwarder to implement exoplayer's new `VideoListener` and attach it in the facade.

### Test(s) added 
I'm not sure there's any worthwhile tests to add here. What we really need is the capability to say, when we've created a player, and that player sends an `onVideoSizeChanged` event, our event listeners all hear it. However, it's not possible to write a test like that right now. It would also be quite an undertaking as we'd need a fake exoplayer which is kept up to date with the version we're targeting.

### Paired with 
No one


core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerVideoRendererEventListener.java
```diff
@@ -1,68 +0,0 @@
...
-    public void onVideoInputFormatChanged(Format format) {
-        for (VideoRendererEventListener listener : listeners) {
-            listener.onVideoInputFormatChanged(format);
-        }
-    }
-
-    @Override
-    public void onDroppedFrames(int count, long elapsedMs) {
-        for (VideoRendererEventListener listener : listeners) {
-            listener.onDroppedFrames(count, elapsedMs);
```
zegnus: we do not have this callback anymore?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/VideoRendererInfoForwarder.java
```diff
@@ -1,91 +0,0 @@
...
-    public void onVideoEnabled(DecoderCounters counters) {
-        HashMap<String, String> callingMethodParameters = new HashMap<>();
-
-        callingMethodParameters.put(Parameters.DECODER_COUNTERS, String.valueOf(counters));
-
-        infoListener.onNewInfo(Methods.ON_VIDEO_ENABLED, callingMethodParameters);
-    }
-
-    @Override
-    public void onVideoDecoderInitialized(String decoderName, long initializedTimestampMs, long initializationDurationMs) {
```
zegnus: we do not have this callback anymore? and others...

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerVideoRendererEventListener.java
joetimmins: No we don't. The only place it is exposed by `exoplayer` is on the `VideoRendererEventListener`, which can't be attached to exoplayer directly any more.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/VideoRendererInfoForwarder.java
Mecharyry: These are exposed through the `analytics listener` now.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerVideoRendererEventListener.java
zegnus: @Mecharyry do we still have this info in the `analytics listener`? or now we cannot know if we have dropped frames at all?

joetimmins: Ah yes, `onDroppedVideoFrames` on exoplayer's AnalyticsListener is the one we want.

PR #170, opened 2018-07-12
Version increase

joetimmins:
## Problem
Need to increase the version number for release. Also, some dependencies were outdated.

## Solution
Update the dependencies and version number.

### Paired with 
Nobody


build.gradle
```diff
@@ -1,9 +1,5 @@
-task wrapper(type: Wrapper) {
```
joetimmins: We don't need this task any more, plus it stuck us to an outdated Gradle version.

core/build.gradle
```diff
@@ -63,8 +63,8 @@ publish {
     version = buildProperties.publish['version'].string
     bintrayUser = buildProperties.publish['bintrayUser'].string
     bintrayKey = buildProperties.publish['bintrayKey'].string
-
+    publishVersion = version
```
joetimmins: `bintray-release` plugin fails without a `publishVersion` in its closure.

PR #172, opened 2018-07-20
Add dropped video frames listener

zegnus:
## Problem

The clients want to know if frames are being dropped

## Solution

Hook up a listener in the exoplayer AnalyticsListener and sufrace it to the NoPlayer interface listeners

### Test(s) added 

No but added a log in the demo activity and tested that it reports correctly

### Screenshots

No UI changes

### Paired with 

Nobody


core/src/main/java/com/novoda/noplayer/Listeners.java
```diff
@@ -130,4 +130,18 @@
      * @param videoSizeChangedListener to remove.
      */
     void removeVideoSizeChangedListener(NoPlayer.VideoSizeChangedListener videoSizeChangedListener);
+
+    /**
+     * Add a given {@link NoPlayer.DroppedVideoFramesListener} to be notified when video playback drop frames
```
Mecharyry: `drops frames`

core/src/main/java/com/novoda/noplayer/NoPlayer.java
```diff
@@ -304,4 +304,9 @@ public void onLoadTimeout() {
 
         void onBeat(NoPlayer player);
     }
+
+    interface DroppedVideoFramesListener {
+
+        void onDroppedVideoFrames(int droppedFrames);
```
Mecharyry: Will clients be measuring time between events themselves?

zegnus: I didn't find the metric particularly useful but I can add it

Mecharyry: It's fine, it was more of a question. Might be better for the client to handle it 😄

PR #175, opened 2018-07-26
Report extensive crashes types

zegnus:
## Problem

No-player wraps exo-player exceptions into big groups, we need crashes with more detail.

## Solution

Create as many error types as exception types exo-player throw

### Test(s) added 

Yes, as much as I could

### Screenshots

No UI changes

### Paired with 

Nobody


core/src/main/java/com/novoda/noplayer/DetailErrorType.java
```diff
@@ -0,0 +1,93 @@
+package com.novoda.noplayer;
+
+/**
+ * Assume all errors are thrown by Exo Player, MEDIA_PLAYER prefix will
+ * indicate that the error is thrown by Meda Player.
```
Mecharyry: Media Player 😄

PR #178, opened 2018-08-31
Allow creating player with custom data source factory for exoplayer

juankysoriano:
## Problem

It is possible that some clients requires for `Exoplayer` and enhanced `DataSource`; for example one can implement a custom caching mechanism or use the `Okhttp` extension that exo player provides: https://github.com/google/ExoPlayer/tree/release-v2/extensions/okhttp

## Solution

Allow `PlayerBuilder` to receive a custom `DataSource.Factory` to be used upon the creation of `MediaSource` at `MediaSourceFactory` 

```
    /**
     * Allows to use a custom data source factory for Exoplayer.
     * e.g A custom data source can implement cachin mechanism
     * e.g The OkHttpDataSource extension can be used
     *
     * @param dataSourceFactory a custom {@link DataSource.Factory} to be used when creating {@link MediaSource}
     * @return {@link PlayerBuilder}
     */
    public PlayerBuilder withDataSourceFactory(DataSource.Factory dataSourceFactory) {
        this.dataSourceFactory = Optional.of(dataSourceFactory);
        return this;
    }
```

### Test(s) added 

There was not tests around the areas affected, no test failed after this

### Paired with 

Nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
+++ b/core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
@@ -4,6 +4,8 @@ import android.content.Context;
 import android.os.Handler;
 import android.os.Looper;
 
+import com.google.android.exoplayer2.source.MediaSource;
+import com.google.android.exoplayer2.upstream.DataSource;
 import com.novoda.noplayer.drm.DownloadedModularDrm;
 import com.novoda.noplayer.drm.DrmHandler;
 import com.novoda.noplayer.drm.DrmType;
@@ -13,6 +15,7 @@ import com.novoda.noplayer.internal.exoplayer.NoPlayerExoPlayerCreator;
 import com.novoda.noplayer.internal.exoplayer.drm.DrmSessionCreatorFactory;
 import com.novoda.noplayer.internal.mediaplayer.NoPlayerMediaPlayerCreator;
 import com.novoda.noplayer.internal.utils.AndroidDeviceVersion;
+import com.novoda.noplayer.internal.utils.Optional;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
```diff
@@ -4,6 +4,8 @@
 import android.os.Handler;
 import android.os.Looper;
 
+import com.google.android.exoplayer2.source.MediaSource;
+import com.google.android.exoplayer2.upstream.DataSource;
```
Mecharyry: ❌ The `no-player` specific classes cannot know about the internals of `exo-player`. `exo-player` classes can only be exposed in either the `com.novoda.noplayer.external.exoplayer` or `com.novoda.noplayer.internal` packages.

core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
```diff
@@ -102,6 +106,19 @@ public PlayerBuilder withDowngradedSecureDecoder() {
         return this;
     }
 
+    /**
+     * Allows to use a custom data source factory for Exoplayer.
+     * e.g A custom data source can implement cachin mechanism
```
Mecharyry: > e.g. A custom data source that can implement a caching mechanism.
> e.g. The OkHttpDataSource extension can be used.

juankysoriano:
But we are exposing exoplayer from the time we do things like `PlayerType.EXO_PLAYER`.

Can you elaborate, given the above, why not? We are already leaking implementation details.

I could try to iterate on that but i am afraid that it is gonna be difficult to avoid doing what we currently do and might have to drop no-player and use exoplayer directly.

Perhaps we can pair later @ry

PR #180, opened 2018-08-31
Add experimental to the build script

Mecharyry:
## Problem
We want to be able to try experimental work on `no-player` without affecting the develop branch. This should allow anyone to create a branch and PR it against an `experimental` branch, on merging this should be released to our private bintray repository. 

## Solution
Add experiment to the build script. The CI work is almost done too.

### Paired with 
@mr-archano helped me sort out the CI for this work.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -24,10 +24,14 @@ buildProperties {
     publish {
         def generateVersion = {
             boolean isSnapshot = cli['bintraySnapshot'].or(false).boolean
+            if (isSnapshot) {
+                return "SNAPSHOT-${System.getenv('BUILD_NUMBER') ?: 'LOCAL'}";
+            }
             boolean isExperimental = cli['bintrayExperimental'].or(false).boolean
-            def snapshotVersion = "SNAPSHOT-${System.getenv('BUILD_NUMBER') ?: 'LOCAL'}"
-            def experimentalVersion = "EXPERIMENTAL-${System.getenv('BUILD_NUMBER') ?: 'LOCAL'}"
-            return isSnapshot ? snapshotVersion : isExperimental ? experimentalVersion : version
+            if (isExperimental) {
+                return "EXPERIMENTAL-${System.getenv('BUILD_NUMBER') ?: 'LOCAL'}";
+            }
+            return version
         }
         using(['version': "${generateVersion()}"])
                 .or(buildProperties.bintray)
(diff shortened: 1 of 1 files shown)


Mecharyry:
test this

core/build.gradle
```diff
@@ -24,7 +24,10 @@ buildProperties {
     publish {
         def generateVersion = {
             boolean isSnapshot = cli['bintraySnapshot'].or(false).boolean
-            return isSnapshot ? "SNAPSHOT-${System.getenv('BUILD_NUMBER')?:'LOCAL'}" : version
+            boolean isExperimental = cli['bintrayExperimental'].or(false).boolean
+            def snapshotVersion = "SNAPSHOT-${System.getenv('BUILD_NUMBER') ?: 'LOCAL'}"
+            def experimentalVersion = "EXPERIMENTAL-${System.getenv('BUILD_NUMBER') ?: 'LOCAL'}"
+            return isSnapshot ? snapshotVersion : isExperimental ? experimentalVersion : version
```
mr-archano: I don't dislike it, but I don't love it either. Pretty sure this can be re-written in a more elegant (aka less confusing) way, just not sure how 😅 
Maybe @devisnik has some idea?

devisnik: Hm, I think it's either what you have or 
````groovy
if (...) 
else if (...)
else 
```

This more verbose control structure might allow you to inline all the local variables, so it might not be that bad in terms of lines of code.
I'm not aware of something in the middle, which I guess you are looking for here. Sorry.

Dorvaryn: What about a when with no arguments ?

```
when {
   isSnapshot -> snapshotVersion 
   isExperimental -> experimentalVersion
   default -> version
} 
```

blundell:
```
if(isSnapshot) {
 return "SNAPSHOT-${System.getenv('BUILD_NUMBER') ?: 'LOCAL'}";
}
boolean isExperimental = cli['bintrayExperimental'].or(false).boolean
if(isExperimental) {
  return "EXPERIMENTAL-${System.getenv('BUILD_NUMBER') ?: 'LOCAL'}";
}
return version;
```

I was just passing through, thought I'd add what I think reads easier

Mecharyry:
Can I haz merge @mr-archano @juankysoriano 😄 I want to try it out 👅

PR #181, opened 2018-08-31
Allow ExoPlayer implementation to be created with a different DataSource.Factory

juankysoriano:
## Problem

Being able to supply an alternative `DataSource` could be beneficial; for example one can implement a custom caching mechanism or use the Okhttp extension that exo player provides: https://github.com/google/ExoPlayer/tree/release-v2/extensions/okhttp

## Solution

Modify the *internals* for creating `NoPlayer` (for exoplayer only) so an `Optional<DataSource.Factory>` with extended behaviour can be supply. 

Then this can be used by advanced clients which implements their own customisation over `NoPlayer`

### Paired with 

@mr-archano and @Dorvaryn helped me on finding a good compromise that doesn't change the client-facing APIs


core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerExoPlayerCreator.java
```diff
@@ -12,14 +13,20 @@
...
 
     private final InternalCreator internalCreator;
 
     public static NoPlayerExoPlayerCreator newInstance(Handler handler) {
-        InternalCreator internalCreator = new InternalCreator(handler);
+        InternalCreator internalCreator = new InternalCreator(handler, Optional.<DataSource.Factory>absent());
+        return new NoPlayerExoPlayerCreator(internalCreator);
+    }
+
+    public static NoPlayerExoPlayerCreator newInstance(Handler handler, Optional<DataSource.Factory> dataSourceFactory) {
```
Dorvaryn: By making this parameter the actual and doing the `optional.of` inside the method we can avoid exposing the internal Optional class to the clients.

PR #184, opened 2018-09-27
Dependency Update and Solo Exo-player

Mecharyry:
We've done a number of changes including: 

#177 Update to ExoPlayer 2.8.4
#180 Add experimental deployment to the build script
#181 & #182 Use a custom datasource factory for exoplayer

We want to release these changes as a patch.


core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ProvisioningModularDrmCallback.java
```diff
@@ -25,7 +28,13 @@
     }
 
     @Override
-    public byte[] executeKeyRequest(UUID uuid, ExoMediaDrm.KeyRequest request) throws Exception {
-        return streamingModularDrm.executeKeyRequest(new ModularDrmKeyRequest(request.getDefaultUrl(), request.getData()));
+    public byte[] executeKeyRequest(UUID uuid, ExoMediaDrm.KeyRequest request, @Nullable String mediaProvidedLicenseServerUrl) throws Exception {
+        String url = request.getDefaultUrl();
+
+        if (TextUtils.isEmpty(url)) {
+            url = mediaProvidedLicenseServerUrl;
```
zegnus: is this ok to be null?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/MediaSourceFactory.java
```diff
@@ -43,6 +46,14 @@ public MediaSource create(Options options,
         }
     }
 
+    private DefaultDataSourceFactory createDataSourceFactory(DefaultBandwidthMeter bandwidthMeter) {
+        if (dataSourceFactory.isPresent()) {
```
zegnus: you can use `map`, `or`... not a blocker

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ProvisioningModularDrmCallback.java
Mecharyry: Rather interestingly I don't think this url is ever used by us. We actually expect a client to pass in the Url as part of the `DrmHandler` because it's a `no-player` class rather than an `ExoPlayer` class.

PR #200, opened 2019-02-25
Allow redirects cross protocols

zegnus:
## Problem

When there is a redirect between `HTTP` and `HTTPS` the request fails

## Solution

- Enable redirects to cross protocols
- Update project dependencies, ExoPlayer to v2.9.6 https://github.com/google/ExoPlayer/blob/release-v2/RELEASENOTES.md#296
- Let the clients pass their own `user-agent` as currently, it had a dummy hardcoded value

### Test(s) added 

No tests added
Tested video playback through the demo application

### Screenshots

No UI changes

### Paired with 

Nobody

### DIFF
--- a/demo/src/main/java/com/novoda/demo/MainActivity.java
+++ b/demo/src/main/java/com/novoda/demo/MainActivity.java
@@ -46,7 +46,9 @@ public class MainActivity extends Activity {
         player = new PlayerBuilder()
                 .withWidevineModularStreamingDrm(drmHandler)
                 .withDowngradedSecureDecoder()
-                .build(this, "Android/Linux");
+                .withUserAgent("Android/Linux")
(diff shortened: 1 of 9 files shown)


demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -46,7 +46,7 @@ protected void onCreate(Bundle savedInstanceState) {
         player = new PlayerBuilder()
                 .withWidevineModularStreamingDrm(drmHandler)
                 .withDowngradedSecureDecoder()
-                .build(this);
+                .build(this, "Android/Linux");
```
zegnus: wait, I'll fix this

zegnus:
closing, I"ll allow the client to decide on this

PR #205, opened 2019-03-22
NO-ADS/Add advert listener

JozefCeluch:
## Problem

We want to allow clients to be notified about advert events

## Solution

This just adds a listener interface and the necessary wiring. It's not used anywhere yet and even the interface is not fully defined yet, both those things will be done in separate PRs.

I also did an unrelated change that moved the `AdsLoader` dependency from the constructor of the `MediaSourceFactory` to its `create` method. This will be useful later on too.

### Test(s) added 

no tests, it's not really doing anything yet

### Screenshots

no UI changes, yet

### Paired with 

nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerExoPlayerCreator.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerExoPlayerCreator.java
@@ -85,7 +85,6 @@ public class NoPlayerExoPlayerCreator {
                     userAgent,
                     handler,
                     dataSourceFactory,
-                    adsLoader,
                     allowCrossProtocolRedirects
             );
 
--- a/core/src/main/java/com/novoda/noplayer/internal/listeners/AdvertListeners.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/listeners/AdvertListeners.java
@@ -1,0 +1,31 @@
+package com.novoda.noplayer.internal.listeners;
+
+import com.novoda.noplayer.NoPlayer;
+
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+class AdvertListeners implements NoPlayer.AdvertListener {
+
+    private final Set<NoPlayer.AdvertListener> listeners = new CopyOnWriteArraySet<>();
+
+    public void add(NoPlayer.AdvertListener listener) {
+        listeners.add(listener);
+    }
+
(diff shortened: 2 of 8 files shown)


core/src/main/java/com/novoda/noplayer/internal/listeners/AdvertListeners.java
```diff
@@ -0,0 +1,31 @@
+package com.novoda.noplayer.internal.listeners;
+
+import com.novoda.noplayer.NoPlayer;
+
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+class AdvertListeners implements NoPlayer.AdvertListener {
+
+    private final Set<NoPlayer.AdvertListener> listeners = new CopyOnWriteArraySet<>();
```
zegnus: 👍

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerExoPlayerCreator.java
```diff
@@ -85,7 +85,6 @@ ExoPlayerTwoImpl create(Context context,
                     userAgent,
                     handler,
                     dataSourceFactory,
-                    adsLoader,
```
juankysoriano: Is the `advertsLoader` still needed as a parameter of the constructor?

juankysoriano: (I can't see why on the diff)

juankysoriano: I am talking about line `74`

JozefCeluch: Yes, the only change is that before we were passing `Optional<AdsLoader>` to both `MediaSourceFactory` and `ExoPlayerFacade` constructors in this method. Now it's only passed to the latter since it doesn't need to be a field in the `MediaSourceFactory`

Mecharyry: Yes. This class is used to create everything related to `no-player`, just the `exo-player` variant. In this case, the `AdsLoader` is passed to the `ExoPlayerFacade` on line 101. The ads loader is then passed to the `MediaSourceFactory` when loading the video, rather than through the constructors.

PR #206, opened 2019-03-25
NO-ADS/Squash Listeners

Mecharyry:
## Problem
We promise this makes sense.

We need to be able to decide whether to notify the video listeners or the advert listeners. At the moment we have a bunch of listeners and forwarders and this is really hard to do. 

## Solution
Have a single place where we can make this decision by introducing a `ExoPlayerListener` which implements all of the interfaces that exo player needs. We'll make the decision and notify either the video or advert listeners in the next PR. 

### Test(s) added 
Updating the existing tests.

### Paired with 
@JozefCeluch


core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerListener.java
```diff
@@ -0,0 +1,694 @@
...
+import com.novoda.noplayer.PlayerState;
+import com.novoda.noplayer.internal.exoplayer.AdvertPlaybackState;
+import com.novoda.noplayer.internal.utils.Optional;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.TimeUnit;
+
+public class ExoPlayerListener implements Player.EventListener, MediaSourceEventListener, AnalyticsListener, VideoListener, DefaultDrmSessionEventListener, AdsLoader {
```
Mecharyry: 😅  all the interfaces...

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerExoPlayerCreator.java
```diff
@@ -104,36 +102,23 @@ ExoPlayerTwoImpl create(Context context,
...
-                    exoPlayerForwarder,
+                    exoPlayerListener,
                     loadTimeout,
                     heart,
                     drmSessionCreator,
                     mediaCodecSelector
             );
         }
-
-        private Optional<AdsLoader> createAdsLoaderFrom(Optional<AdvertsLoader> advertsLoader) {
```
zegnus: somebody else is going to do this then?

Mecharyry: Somebody already does this 😄

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/MediaSourceFactory.java
```diff
@@ -43,19 +42,18 @@ public MediaSourceFactory(Context context,
...
+                              ExoPlayerListener exoPlayerListener,
+                              DefaultBandwidthMeter bandwidthMeter) {
         DefaultDataSourceFactory defaultDataSourceFactory = createDataSourceFactory(bandwidthMeter);
 
         MediaSource contentMediaSource = getMediaSourceFor(options, uri, defaultDataSourceFactory);
 
-        if (advertsLoader.isPresent()) {
-            AdsMediaSource adsMediaSource = new AdsMediaSource(contentMediaSource, defaultDataSourceFactory, advertsLoader.get(), null);
-            adsMediaSource.addEventListener(handler, mediaSourceEventListener);
+        if (exoPlayerListener.canLoadAdverts()) {
```
mr-archano: This is looking weird to me, as I wouldn't expect a listener (by definition passive) to dictate changes in behaviour here.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/MediaSourceFactory.java
```diff
@@ -43,19 +42,18 @@ public MediaSourceFactory(Context context,
...
+            AdsMediaSource adsMediaSource = new AdsMediaSource(contentMediaSource, defaultDataSourceFactory, exoPlayerListener, null);
```
mr-archano: As far I understand the `AdsMediaSource` here is decorating the `contentMediasource`, but I don't fully grasp why you need to add listener when creating one.  Shouldn't be responsibility of the consumer deciding whether they are interested in adverts (so to wrap the original source) and then subscribing to those additional events too?

PR #207, opened 2019-03-27
Bind AdvertsListener directly to the AdsLoader

Mecharyry:
## Problem
Had a discussion with @mr-archano and we were both a little unhappy with how #206 the adverts loading was part of the squashed `ExoPlayerListener` class. 

## Solution
**This is an alternative feature branch where everything has not been collapsed.**

Instead the advert listener is bound to the `ExoPlayerForwarder` but then we grab this `AdvertListener` and bind it to the `AdvertsLoader`. The `AdvertsLoader` still implements the `AdsLoader` and `Player.EventListener` interfaces, we listen to what we want and forward when necessary to the `AdvertListener`. At the moment this is just the load of the advert.

The only issue is that we have to expose the `class` rather than the `interface` of the `AdvertsLoader` so that we can bind the `AdvertsListener` when it becomes available. 

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerForwarder.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerForwarder.java
@@ -14,6 +14,7 @@ public class ExoPlayerForwarder {
     private final NoPlayerAnalyticsListener analyticsListener;
     private final ExoPlayerVideoListener videoListener;
     private final ExoPlayerDrmSessionEventListener drmSessionEventListener;
+    private NoPlayer.AdvertListener advertListeners;
 
     public ExoPlayerForwarder() {
         exoPlayerEventListener = new EventListener();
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerForwarder.java
```diff
@@ -43,6 +44,10 @@ public AnalyticsListener analyticsListener() {
         return analyticsListener;
     }
 
+    public NoPlayer.AdvertListener advertListener() {
```
mr-archano: Shouldn't this be `@Nullable` then? (or is `Optional` preferred in the project?)

Mecharyry: Ah yes 😄

Mecharyry:
`NoPlayerAdsLoader` was already defined in an earlier PR @mr-archano, we just removed it in the squash PR from yesterday 😄

PR #208, opened 2019-03-27
NO-ADS/Advert position and duration

JozefCeluch:
## Problem

We need to expose advert break duration and playhead position within the advert group. 

## Solution

We need to calculate the duration of the ads inside the `AdGroup` manually. We achieve this by combining the durations of all adverts in the group that is being currently played. To calculate the progress we combine the lengths of all the finished adverts in the group and then add current ad progress to it.

There is a bit of a weird behavior in exoplayer where the advert durations go from a known value to unknown for a brief period (seems like it's until they get cached or something similar). In order to prevent the durations to jump we expose the `NoPlayerAdsLoader` to the facade so that it can fallback to the advert duration values there in those cases.

### Test(s) added 

a bunch of tests to test the new behavior, the setup in the tests exposes the internal implementation details quite a bit but I'm not sure how to make them better

### Screenshots
no UI changes

### Paired with 

@Mecharyry

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
@@ -2,6 +2,7 @@ package com.novoda.noplayer.internal.exoplayer;
 
 import android.net.Uri;
 import android.support.annotation.Nullable;
+
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.Player;
 import com.google.android.exoplayer2.SimpleExoPlayer;
@@ -25,6 +26,8 @@ import com.novoda.noplayer.model.PlayerVideoTrack;
 
 import java.util.List;
 
+// Not much we can do, wrapping ExoPlayer is a lot of work
+@SuppressWarnings("PMD.GodClass")
 class ExoPlayerFacade {
 
     private static final boolean RESET_POSITION = true;
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
@@ -87,7 +87,6 @@ class ExoPlayerFacade {
         if (!isPlayingAdvert() || !(currentTimeline instanceof SinglePeriodAdTimeline) || adsLoader.isAbsent()) {
             return 0;
         }
-        NoPlayerAdsLoader loader = adsLoader.get();
         SinglePeriodAdTimeline adTimeline = ((SinglePeriodAdTimeline) currentTimeline);
         Timeline.Period period = adTimeline.getPeriod(0, new Timeline.Period());
         int currentAdGroupIndex = exoPlayer.getCurrentAdGroupIndex();
@@ -97,12 +96,35 @@ class ExoPlayerFacade {
         for (int i = 0; i < numberOfAdverts; i++) {
             long advertDurationInMicros = period.getAdDurationUs(currentAdGroupIndex, i);
             if (advertDurationInMicros == C.TIME_UNSET) {
-                advertDurationInMicros = loader.advertDurationBy(currentAdGroupIndex, i);
+                advertDurationInMicros = adsLoader.get().advertDurationBy(currentAdGroupIndex, i);
             }
             advertBreakDurationInMicros += advertDurationInMicros;
         }
 
         return C.usToMs(advertBreakDurationInMicros);
+    }
+
+    long positionInAdvertBreakInMillis() {
+        assertVideoLoaded();
+        Timeline currentTimeline = exoPlayer.getCurrentTimeline();
+        if (!isPlayingAdvert() || !(currentTimeline instanceof SinglePeriodAdTimeline) || adsLoader.isAbsent()) {
+            return 0;
+        }
+
+        SinglePeriodAdTimeline adTimeline = ((SinglePeriodAdTimeline) currentTimeline);
+        Timeline.Period period = adTimeline.getPeriod(0, new Timeline.Period());
+        int currentAdGroupIndex = exoPlayer.getCurrentAdGroupIndex();
+        int previousAdIndexInGroup = exoPlayer.getCurrentAdIndexInAdGroup() - 1;
+
+        long playedAdvertBreakDurationInMicros = 0;
+        for (int i = previousAdIndexInGroup; i >= 0; i--) {
+            long advertDurationInMicros = period.getAdDurationUs(currentAdGroupIndex, i);
+            if (advertDurationInMicros == C.TIME_UNSET) {
+                advertDurationInMicros = adsLoader.get().advertDurationBy(currentAdGroupIndex, i);
+            }
+            playedAdvertBreakDurationInMicros += advertDurationInMicros;
+        }
+        return C.usToMs(playedAdvertBreakDurationInMicros) + playheadPositionInMillis();
     }
 
     long mediaDurationInMillis() throws IllegalStateException {
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
@@ -119,10 +119,11 @@ class ExoPlayerFacade {
     private long combinedAdvertDurationInGroup(Timeline.Period period, int numberOfAdvertsToInclude) {
         int adGroupIndex = exoPlayer.getCurrentAdGroupIndex();
         long advertBreakDurationInMicros = 0;
+        NoPlayerAdsLoader noPlayerAdsLoader = adsLoader.get();
         for (int i = 0; i < numberOfAdvertsToInclude; i++) {
             long advertDurationInMicros = period.getAdDurationUs(adGroupIndex, i);
             if (advertDurationInMicros == C.TIME_UNSET) {
-                advertDurationInMicros = adsLoader.get().advertDurationBy(adGroupIndex, i);
+                advertDurationInMicros = noPlayerAdsLoader.advertDurationBy(adGroupIndex, i);
             }
             advertBreakDurationInMicros += advertDurationInMicros;
         }
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
@@ -120,6 +120,7 @@ class ExoPlayerFacade {
         int adGroupIndex = exoPlayer.getCurrentAdGroupIndex();
         long advertBreakDurationInMicros = 0;
         NoPlayerAdsLoader noPlayerAdsLoader = adsLoader.get();
+
         for (int i = 0; i < numberOfAdvertsToInclude; i++) {
             long advertDurationInMicros = period.getAdDurationUs(adGroupIndex, i);
             if (advertDurationInMicros == C.TIME_UNSET) {
@@ -127,6 +128,7 @@ class ExoPlayerFacade {
             }
             advertBreakDurationInMicros += advertDurationInMicros;
         }
+
         return advertBreakDurationInMicros;
     }
 
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
@@ -91,32 +91,34 @@ class ExoPlayerFacade {
     long advertBreakDurationInMillis() throws IllegalStateException {
         assertVideoLoaded();
         Timeline currentTimeline = exoPlayer.getCurrentTimeline();
-        if (!isPlayingAdvert() || adsLoader.isAbsent() || !(currentTimeline instanceof SinglePeriodAdTimeline)) {
-            return 0;
+        if (isPlayingAdvert() && adsLoader.isPresent() && currentTimeline instanceof SinglePeriodAdTimeline) {
+            SinglePeriodAdTimeline adTimeline = (SinglePeriodAdTimeline) currentTimeline;
+            Timeline.Period period = adTimeline.getPeriod(0, new Timeline.Period());
+
+            int currentAdGroupIndex = exoPlayer.getCurrentAdGroupIndex();
+            int advertCount = period.getAdCountInAdGroup(currentAdGroupIndex);
+
+            long advertBreakDurationInMicros = combinedAdvertDurationInGroup(period, advertCount);
+            return C.usToMs(advertBreakDurationInMicros);
         }
-        SinglePeriodAdTimeline adTimeline = (SinglePeriodAdTimeline) currentTimeline;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -66,14 +70,67 @@ boolean isPlaying() {
...
+
     long playheadPositionInMillis() throws IllegalStateException {
         assertVideoLoaded();
         return exoPlayer.getCurrentPosition();
     }
 
-    long mediaDurationInMillis() throws IllegalStateException {
+    long advertBreakDurationInMillis() {
         assertVideoLoaded();
-        return exoPlayer.getDuration();
```
Mecharyry: This removal diff is not confusing at all 😂 it is still there, stupid GitHub 😂

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -66,14 +70,67 @@ boolean isPlaying() {
...
+    }
+
     long playheadPositionInMillis() throws IllegalStateException {
         assertVideoLoaded();
         return exoPlayer.getCurrentPosition();
     }
 
-    long mediaDurationInMillis() throws IllegalStateException {
+    long advertBreakDurationInMillis() {
         assertVideoLoaded();
```
zegnus: should this method also throw an `IllegalStateException` because of this line?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -66,14 +70,67 @@ boolean isPlaying() {
...
+
+        int currentAdGroupIndex = exoPlayer.getCurrentAdGroupIndex();
+        int advertCount = period.getAdCountInAdGroup(currentAdGroupIndex);
+
+        long advertBreakDurationInMicros = combinedAdvertDurationInGroup(period, advertCount);
+        return C.usToMs(advertBreakDurationInMicros);
+    }
+
+    long positionInAdvertBreakInMillis() {
+        assertVideoLoaded();
```
zegnus: same here

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -66,14 +70,67 @@ boolean isPlaying() {
...
+        SinglePeriodAdTimeline adTimeline = (SinglePeriodAdTimeline) currentTimeline;
+        Timeline.Period period = adTimeline.getPeriod(0, new Timeline.Period());
+
+        int advertCount = exoPlayer.getCurrentAdIndexInAdGroup();
+
+        long playedAdvertBreakDurationInMicros = combinedAdvertDurationInGroup(period, advertCount);
+        return C.usToMs(playedAdvertBreakDurationInMicros) + playheadPositionInMillis();
+    }
+
+    private long combinedAdvertDurationInGroup(Timeline.Period period,
```
zegnus: fits in one line

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -66,14 +70,67 @@ boolean isPlaying() {
...
+                                               int numberOfAdvertsToInclude) {
+        int adGroupIndex = exoPlayer.getCurrentAdGroupIndex();
+        long advertBreakDurationInMicros = 0;
+        for (int i = 0; i < numberOfAdvertsToInclude; i++) {
+            long advertDurationInMicros = period.getAdDurationUs(adGroupIndex, i);
+            if (advertDurationInMicros == C.TIME_UNSET) {
+                advertDurationInMicros = adsLoader.get().advertDurationBy(adGroupIndex, i);
```
zegnus: can `adsLoader.get()` be extracted outside of the loop? so that its done only once

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -32,6 +33,8 @@
     private AdvertsLoader.Cancellable loadingAds;
 
     private Optional<NoPlayer.AdvertListener> advertListener = Optional.absent();
+    @Nullable
+    private List<AdvertBreak> advertBreaks;
```
zegnus: I guess you already thought of this but, can we have an empty list instead?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -67,9 +70,10 @@ public void start(final EventListener eventListener, AdViewProvider adViewProvid
 
     private final AdvertsLoader.Callback advertsLoadedCallback = new AdvertsLoader.Callback() {
         @Override
-        public void onAdvertsLoaded(List<AdvertBreak> advertBreaks) {
+        public void onAdvertsLoaded(List<AdvertBreak> breaks) {
             loadingAds = null;
-            adPlaybackState = AdvertPlaybackState.from(advertBreaks);
+            advertBreaks = breaks;
```
zegnus: can we instead `advertBreaks.clear(); advertBreaks.addAll(breaks)` so that the local reference stays in place

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -92,6 +96,19 @@ private void updateAdPlaybackState() {
...
     }
 
+    public long advertDurationBy(int advertGroupIndex, int advertIndexInAdvertGroup) {
+        if (advertBreaks == null || advertGroupIndex >= advertBreaks.size()) {
+            return 0;
+        }
+
+        AdvertBreak advertBreak = advertBreaks.get(advertGroupIndex);
+        if (advertIndexInAdvertGroup >= advertBreak.adverts().size()) {
+            return 0;
```
zegnus: Looks like these two cases is actually an error, shall we throw an exception then? or `Log.e(...)`?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
Mecharyry: Yes 😅

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
Mecharyry: Same 😅

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
Mecharyry: Good point 💯

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: Yes we probably can 😄

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: And the field could then be final. 👍

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: Yes we probably could, it would be a fundamental flaw in ExoPlayer or our code if this happened.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
JozefCeluch: https://github.com/novoda/no-player/pull/208/commits/86e413a0931385ca3438337b8df2f86fa13f64c1

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
JozefCeluch: https://github.com/novoda/no-player/pull/208/commits/86e413a0931385ca3438337b8df2f86fa13f64c1

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
JozefCeluch: https://github.com/novoda/no-player/pull/208/commits/bc9549fec4b44fc13816f97e8cad00ca8a00b009

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
JozefCeluch: https://github.com/novoda/no-player/pull/208/commits/65073331f1fc734fe2801ed4da75a899056deab3

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
JozefCeluch: https://github.com/novoda/no-player/pull/208/commits/7ddefbc2641359663aa5fcde7836be5e3c08ad31

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
JozefCeluch: https://github.com/novoda/no-player/pull/208/commits/7ddefbc2641359663aa5fcde7836be5e3c08ad31

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
JozefCeluch: Throwing our runtime exception would be exactly the same behavior as if we didn't do any checks at all and we would just get the items from the lists directly. So wouldn't it be better to remove the checks and make it simpler? (Since this case should never happen)

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -67,9 +70,11 @@ public void start(final EventListener eventListener, AdViewProvider adViewProvid
 
     private final AdvertsLoader.Callback advertsLoadedCallback = new AdvertsLoader.Callback() {
         @Override
-        public void onAdvertsLoaded(List<AdvertBreak> advertBreaks) {
+        public void onAdvertsLoaded(List<AdvertBreak> breaks) {
             loadingAds = null;
-            adPlaybackState = AdvertPlaybackState.from(advertBreaks);
+            advertBreaks.clear();
+            advertBreaks.addAll(breaks);
```
pablisco: If we make `advertBreaks` non final then we can have an empty list to start with (`Collections.emptyList()`) and here we can make a copy of the provided list and reassign it (`advertBreaks  = new ArrayList<>(list)`).
Which would make it more thread safe

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -66,14 +70,69 @@ boolean isPlaying() {
...
+        int adGroupIndex = exoPlayer.getCurrentAdGroupIndex();
+        long advertBreakDurationInMicros = 0;
+        NoPlayerAdsLoader noPlayerAdsLoader = adsLoader.get();
+
+        for (int i = 0; i < numberOfAdvertsToInclude; i++) {
+            long advertDurationInMicros = period.getAdDurationUs(adGroupIndex, i);
+            if (advertDurationInMicros == C.TIME_UNSET) {
+                advertDurationInMicros = noPlayerAdsLoader.advertDurationBy(adGroupIndex, i);
+            }
+            advertBreakDurationInMicros += advertDurationInMicros;
```
pablisco: We can split this into two parts with an `if-else` and avoid reassignment:
```java
if (advertDurationInMicros == C.TIME_UNSET) {
  advertBreakDurationInMicros += noPlayerAdsLoader.advertDurationBy(adGroupIndex, i);
} else {
  advertBreakDurationInMicros += period.getAdDurationUs(adGroupIndex, i);
}
```

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -66,14 +70,69 @@ boolean isPlaying() {
...
+        if (!isPlayingAdvert() || adsLoader.isAbsent() || !(currentTimeline instanceof SinglePeriodAdTimeline)) {
+            return 0;
+        }
+        SinglePeriodAdTimeline adTimeline = (SinglePeriodAdTimeline) currentTimeline;
+        Timeline.Period period = adTimeline.getPeriod(0, new Timeline.Period());
+
+        int advertCount = exoPlayer.getCurrentAdIndexInAdGroup();
+
+        long playedAdvertBreakDurationInMicros = combinedAdvertDurationInGroup(period, advertCount);
+        return C.usToMs(playedAdvertBreakDurationInMicros) + playheadPositionInMillis();
```
pablisco: If we extract everything after the early return into a function we can invert the if statement to be positive, easier to read than having negations IMHO

pablisco: Same for the previous method

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
Mecharyry: Am I missing something? We'd need to do two calls to `period.getAdDurationUs`, which is why we have the reassignment.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: f76ddba

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
Mecharyry: 4341fb5

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
pablisco: Ah! true, my bad, I thought the condition was not based on that value 🤐

PR #209, opened 2019-03-29
NO-ADS/Advert Start End Complete

Mecharyry:
## Problem
- We want to be able to notify clients of when advert break start and end events occurs, as well as advert start and end events. 

## Solution
- Define the public API for receiving the above events. This mainly involved emitting the given event, start advert break for example, with an identifier specified by the client. We introduced the appropriate methods on the listeners and created the additional id models. 
- Determine when the events are emitted by `exo-player` and forward the appropriate event to the client with the associated identifier. 

**Then we manually tested, at which point we discovered:**

- The ordering of the advert breaks between this PR and #208 was lost. This ordering is extremely important as `exo-player` uses the order for the method calls to `currentAdvertGroup` and `currentAdvert`. We've introduced a `AdvertPlaybackState` model to wrap the `AdvertBreaks` and the `AdPlaybackState` to ensure the ordering is preserved between loading adverts and passing them to the `exo-player`. 
- The `Optional<AdvertListener>` is tedious to use because we can't add lambdas and we do not have any useful functions on the `Optional` interface. We tried adding some more useful methods to the `Optional` such as a `ifPresent(Func1)` but actually it didn't improve readability at all, due to the lack of any lambdas. In the end we opted to add the `NoOp` version of the advert listener and replace it with calls to `bind`. 

### Test(s) added 
We didn't actually add any tests in this PR. I'm going to open a PR after this one that goes through and adds tests to the `NoPlayerAdsLoader`. 

### Paired with 
@JozefCeluch

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
@@ -161,10 +161,8 @@ public class NoPlayerAdsLoader implements AdsLoader, Player.EventListener {
             adGroupIndex = player.getCurrentAdGroupIndex();
             adIndexInGroup = player.getCurrentAdIndexInAdGroup();
 
-            if (reason == Player.TIMELINE_CHANGE_REASON_PREPARED) {
-                if (adGroupIndex != -1 && adIndexInGroup != -1) {
-                    notifyAdvertStart(advertBreaks.get(adGroupIndex));
-                }
+            if (reason == Player.TIMELINE_CHANGE_REASON_PREPARED && adGroupIndex != -1 && adIndexInGroup != -1) {
+                notifyAdvertStart(advertBreaks.get(adGroupIndex));
             }
         }
     }
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
@@ -162,7 +162,9 @@ public class NoPlayerAdsLoader implements AdsLoader, Player.EventListener {
             adIndexInGroup = player.getCurrentAdIndexInAdGroup();
 
             if (reason == Player.TIMELINE_CHANGE_REASON_PREPARED) {
-                notifyAdvertStart();
+                if (adGroupIndex != -1 && adIndexInGroup != -1) {
+                    notifyAdvertStart(advertBreaks.get(adGroupIndex));
+                }
             }
         }
     }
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -154,25 +160,71 @@ public void onTimelineChanged(Timeline timeline, @Nullable Object manifest, int
...
             adIndexInGroup = player.getCurrentAdIndexInAdGroup();
+            notifyAdvertStart();
         }
     }
 
-    private void notifyEventIfPossible(String event) {
-        if (advertListener.isPresent()) {
-            advertListener.get().onAdvertEvent(event);
+    private void notifyAdvertEnd() {
+        AdvertBreak advertBreak = advertBreaks.get(adGroupIndex);
```
mr-archano: I noticed that this is happening in 2 places (here and in `notifyAdverStart()`: wouldn't be better to pass the `advertBreak` as parameter then?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -154,25 +160,71 @@ public void onTimelineChanged(Timeline timeline, @Nullable Object manifest, int
...
+        List<Advert> adverts = advertBreak.adverts();
+        advertListener.onAdvertEnd(adverts.get(adIndexInGroup).advertId());
+
+        if (adIndexInGroup == adverts.size() - 1) {
+            advertListener.onAdvertBreakEnd(advertBreak.advertBreakId());
+        }
+    }
+
+    private void notifyAdvertStart() {
+        if (adGroupIndex != -1 && adIndexInGroup != -1) {
```
mr-archano: for `notifyAdvertEnd()` this check is external: worth keeping this outside too?

core/src/main/java/com/novoda/noplayer/internal/listeners/AdvertListeners.java
```diff
@@ -22,10 +24,30 @@ public void clear() {
     }
 
     @Override
-    public void onAdvertEvent(String event) {
+    public void onAdvertBreakStart(AdvertBreakId advertBreakId) {
         for (NoPlayer.AdvertListener listener : listeners) {
-            listener.onAdvertEvent(event);
+            listener.onAdvertBreakStart(advertBreakId);
```
danybony: Did you think about renaming this whole class (I cannot comment there) to `CompositeAdvertListener`, to make it more explicit what it's doing?

Mecharyry: I don't really follow? Oh because it deals with a group of advert listeners? tbh, this is the project convention, we always do the group as `Listeners` and the single as `Listener`. We could open an issue to update the naming if you think it is confusing though? It is pretty easy to occasionally miss that `s` 😂

Mecharyry:
**tldr;** `AdvertPlaybackState` is just a wrapper around the `exo-player` `AdPlaybackState` and the client `List<AdvertBreak>`

@danybony the `AdvertPlaybackState` is a wrapper around `List<AdvertBreak>` and `AdPlaybackState`. `AdPlaybackState` is owned by `exo-player` and represents the state it needs to know in order to play an advert. The `AdPlaybackState` is created from the client `AdvertBreak`s, from which we get the necessary information. One of the unfortunate things is that the `List<AdvertBreak>` needs to be ordered correctly for the `AdPlaybackState` to work correctly, this means that we also need to pass this back to the `NoPlayerAdsLoader` that caches them to ensure that the order matches the `exo-player` version. Super annoying but completely necessary 😄

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: 55a7212

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: 55a7212

core/src/main/java/com/novoda/noplayer/internal/listeners/AdvertListeners.java
danybony: Yeah, sometimes I forget about this project convention and I find `Composite` easier to understand, as it is a known pattern.
Good to be addressed outside this PR 👍

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -154,25 +160,70 @@ public void onTimelineChanged(Timeline timeline, @Nullable Object manifest, int
         if (player != null) {
             adGroupIndex = player.getCurrentAdGroupIndex();
             adIndexInGroup = player.getCurrentAdIndexInAdGroup();
+
+            if (reason == Player.TIMELINE_CHANGE_REASON_PREPARED && adGroupIndex != -1 && adIndexInGroup != -1) {
```
mr-archano: not a blocker: but what about extracting a method for this check?
```
adGroupIndex != -1 && adIndexInGroup != -1
```
it's used in few places and I wonder if we could make it more understandable while reducing some duplication

PR #210, opened 2019-03-29
NO-ADS/Tests for ads loader

JozefCeluch:
## Problem

Some tests were missing.

## Solution

Add all the tests (for the `NoPlayerAdsLoader`, that is).

### Test(s) added 

All the tests were added, kinda. I'm not sure if it's worth it adding the tests for the behavior where we would need to set up a lot of preconditions in all mocks. 

### Screenshots
no uI

### Paired with 

nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackState.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackState.java
@@ -5,6 +5,7 @@ import com.google.android.exoplayer2.source.ads.AdPlaybackState;
 import com.novoda.noplayer.Advert;
 import com.novoda.noplayer.AdvertBreak;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
@@ -14,16 +15,16 @@ final class AdvertPlaybackState {
     private final List<AdvertBreak> advertBreaks;
 
     static AdvertPlaybackState from(List<AdvertBreak> advertBreaks) {
-        Collections.sort(advertBreaks, new AdvertBreakStartTimeComparator());
+        List<AdvertBreak> sortedAdvertBreaks = sortAdvertBreaksByStartTime(advertBreaks);
 
-        long[] advertOffsets = advertBreakOffset(advertBreaks);
+        long[] advertOffsets = advertBreakOffset(sortedAdvertBreaks);
         AdPlaybackState adPlaybackState = new AdPlaybackState(advertOffsets);
 
-        int advertBreaksCount = advertBreaks.size();
+        int advertBreaksCount = sortedAdvertBreaks.size();
         long[][] advertBreaksWithAdvertDurations = new long[advertBreaksCount][];
 
         for (int i = 0; i < advertBreaksCount; i++) {
-            AdvertBreak advertBreak = advertBreaks.get(i);
+            AdvertBreak advertBreak = sortedAdvertBreaks.get(i);
             List<Advert> adverts = advertBreak.adverts();
 
             int advertsCount = adverts.size();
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackState.java
```diff
@@ -57,6 +58,12 @@ AdPlaybackState adPlaybackState() {
         return advertBreaks;
     }
 
+    private static List<AdvertBreak> sortAdvertBreaksByStartTime(List<AdvertBreak> advertBreaks) {
+        List<AdvertBreak> sortedAdvertBreaks = new ArrayList<>(advertBreaks);
```
Mecharyry: We're doing this here and in the `NoPlayerAdsLoader`, do we want to make a single copy and use that?

JozefCeluch: I guess we don't need to make another copy in the ads loader 👍

core/src/test/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoaderTest.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Arrays;
+
+import static com.novoda.noplayer.AdvertBreakFixtures.anAdvertBreak;
+import static com.novoda.noplayer.AdvertFixtures.anAdvert;
+import static org.fest.assertions.api.Assertions.assertThat;
+import static org.mockito.BDDMockito.then;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+public class NoPlayerAdsLoaderTest {
```
Mecharyry: I guess we can't really test anything using the `exo-player` listeners? Or do you think that goes too much into implementation details without going for the behaviour that we care about?

JozefCeluch: I tried and it seemed a bit forced because it would look like this:
call `adsLoader.bind(advertListener)`
call `adsLoade.start()`
capture `advertsLoadedCallback`
call `advertsLoadedCallback.onAdvertsLoaded()` with correct adverts
setup the `player` so it returns reasonable values

call `adsLoader.onPositionDiscontinuity()` or `adsLoader.onTimelineChanged()` with correct parameters 

assert that correct `advertListener` method was called

😅

Mecharyry: Yea I figured it might look a bit weird 😂 fair enough.

core/src/test/java/com/novoda/noplayer/AdvertFixtures.java
```diff
@@ -0,0 +1,39 @@
+package com.novoda.noplayer;
+
+import android.net.Uri;
+
+import static org.mockito.Mockito.mock;
+
+public class AdvertFixtures {
```
mr-archano: 💯

core/src/test/java/com/novoda/noplayer/AdvertBreakFixtures.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Collections;
+import java.util.List;
+
+public class AdvertBreakFixtures {
+
+    private AdvertBreakId advertBreakId = new AdvertBreakId("advert_break_id");
+    private long startTimeInMillis = 1000;
+    private List<Advert> adverts = new ArrayList<>();
+
+    public static AdvertBreakFixtures anAdvertBreak() {
```
mr-archano: 💯

PR #212, opened 2019-04-05
NO-ADS/Skipping adverts based on start position

Mecharyry:
## Problem
We want to be able to skip adverts based on the initial playback position. In this case if the initial position also happens to be an advert position we play the advert first.

## Solution
Create a `SkippedAdverts` collaborator that determines the adverts to skip based on the `initial position` the `client model (AdvertBreaks)` and the `adPlaybackState`, that we will modify. 

### Test(s) added 
Yes, tested the `SkippedAdverts` collaborator. Unfortunately it has somewhat of a hard dependency on the `AdvertPlaybackState` working so if that fails so will these tests. I couldn't really avoid it without redoing a lot of work that class already does. 

#### Manual testing

Case | outcome
--- | ---
What happens if you pass a time that is on an advert? | Made it so that this advert will play before the main content. 
What happens if you perform a seek forwards? | It will not skip adverts, the player needs to be in the preparing state for the timeline and calling the `onTimelineChanged` in the `advertsLoader`. As far as I know, this can never happen except on initial load. 
What happens if you perform a seek backwards? | Nothing, the adverts have already been skipped, they will not trigger again.

## Screen Capture
Just shows that loading the video passing a time skips the adverts before that point.

skipping_adverts mp4

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
@@ -3,6 +3,7 @@ package com.novoda.noplayer.internal.exoplayer;
 import android.os.Handler;
 import android.os.Looper;
 import android.support.annotation.Nullable;
+import android.util.Log;
 
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.Player;
@@ -156,11 +157,20 @@ public class NoPlayerAdsLoader implements AdsLoader, Player.EventListener {
             return;
         }
 
-        if (player != null) {
+        if (player != null && adPlaybackState != null) {
             adGroupIndex = player.getCurrentAdGroupIndex();
             adIndexInGroup = player.getCurrentAdIndexInAdGroup();
+            long contentPosition = player.getContentPosition();
 
             if (reason == Player.TIMELINE_CHANGE_REASON_PREPARED && isPlayingAdvert()) {
+                if (contentPosition > 0) {
+                    for (int i = adGroupIndex; i >= 0; i--) {
+                        adPlaybackState = adPlaybackState.withSkippedAdGroup(i);
+                    }
+                    updateAdPlaybackState();
+                    return;
+                }
+
                 notifyAdvertStart(advertBreaks.get(adGroupIndex));
             }
         }
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
@@ -3,7 +3,6 @@ package com.novoda.noplayer.internal.exoplayer;
 import android.os.Handler;
 import android.os.Looper;
 import android.support.annotation.Nullable;
-import android.util.Log;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -156,11 +156,18 @@ public void onTimelineChanged(Timeline timeline, @Nullable Object manifest, int
             return;
         }
 
-        if (player != null) {
+        if (player != null && adPlaybackState != null) {
```
mr-archano: without knowing much of how this used to work - it looks like now the check is stricter than before: how does this additional condition influence things like L171 (`notifyAdvertStart(advertBreaks.get(adGroupIndex));`)?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SkippedAdverts.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.List;
+
+final class SkippedAdverts {
+
+    private SkippedAdverts() {
+        // Utility class.
+    }
+
+    static AdPlaybackState from(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
+        AdPlaybackState advertPlaybackState = adPlaybackState;
```
mr-archano: `nit:` what about `adPlaybackStateWithSkippedAdGroups`?

core/src/test/java/com/novoda/noplayer/internal/exoplayer/SkippedAdvertsTest.java
```diff
@@ -0,0 +1,77 @@
...
+    private static final AdvertBreak FOURTH_ADVERT_BREAK = anAdvertBreak()
+            .withStartTimeInMillis(FORTY_SECONDS_IN_MILLIS)
+            .withAdvert(anAdvert().build())
+            .build();
+
+    private final AdvertPlaybackState advertPlaybackState = AdvertPlaybackState.from(Arrays.asList(FIRST_ADVERT_BREAK, SECOND_ADVERT_BREAK, THIRD_ADVERT_BREAK, FOURTH_ADVERT_BREAK));
+
+    @Test
+    public void doesNotSkipAdvert_whenCurrentPositionIsAtAdvertPosition() {
+        AdPlaybackState adPlaybackState = SkippedAdverts.from(THIRTY_SECONDS_IN_MILLIS, advertPlaybackState.advertBreaks(), advertPlaybackState.adPlaybackState());
```
mr-archano: `nit:` it seems like we really don't care about `advertPlaybackState`, what if we extracted constants for advert breaks and `AdPlaybackState`?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: Interesting point. So this shouldn't actually affect anything and in retrospect I believe we should have updated some of the other checks to also include this. My rationale is that this `AdPlaybackState` would only be `null` if the adverts failed to load or this class has already been released, in both cases we shouldn't be performing a notification for adverts anyway.

PR #213, opened 2019-04-05
Add max video bitrate

frapontillo:
## Problem

There might be the need to set a maximum video bitrate in order to manually limit the quality of the video in a forceful way (e.g. SD only), even if the manifest publishes higher quality representations.

## Solution

Use `DefaultTrackSelector`'s parameters to set a maximum video bitrate both when we `loadVideo` and at any moment during the playing of the video.

### Test(s) added 

Yes, just to make sure we delegate the setting and clearing of the maximum video bitrate.

### Paired with 

Checked for the new API with @Mecharyry

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelectorCreator.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelectorCreator.java
@@ -25,6 +25,10 @@ class CompositeTrackSelectorCreator {
                 Clock.DEFAULT
         );
         DefaultTrackSelector trackSelector = new DefaultTrackSelector(adaptiveTrackSelectionFactory);
+        DefaultTrackSelector.Parameters trackSelectorParameters = trackSelector.buildUponParameters()
+                .setMaxVideoBitrate(options.maxVideoBitrate())
+                .build();
+        trackSelector.setParameters(trackSelectorParameters);
 
         ExoPlayerTrackSelector exoPlayerTrackSelector = ExoPlayerTrackSelector.newInstance(trackSelector);
         ExoPlayerAudioTrackSelector audioTrackSelector = new ExoPlayerAudioTrackSelector(exoPlayerTrackSelector);
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/ExoPlayerTrackSelector.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/ExoPlayerTrackSelector.java
@@ -80,4 +80,20 @@ public class ExoPlayerTrackSelector {
                 && trackGroups.get(groupIndex).length > 0
                 && trackInfo().getAdaptiveSupport(audioRendererIndex.get(), groupIndex, false) != RendererCapabilities.ADAPTIVE_NOT_SUPPORTED;
     }
+
+    void clearMaxVideoBitrate() {
+        setMaxVideoBitrateParameter(Integer.MAX_VALUE);
+    }
+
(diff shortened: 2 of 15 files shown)


frapontillo:
ok to test

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -17,14 +19,16 @@
 
 public class MainActivity extends Activity {
 
-    private static final String URI_VIDEO_WIDEVINE_EXAMPLE_MODULAR_MPD = "https://storage.googleapis.com/content-samples/multi-audio/manifest.mpd";
+    private static final String URI_VIDEO_WIDEVINE_EXAMPLE_MODULAR_MPD = "https://storage.googleapis.com/wvmedia/clear/h264/tears/tears.mpd";
```
frapontillo: I had to change this since the MPD we were using only published one video `Representation`.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelectorCreator.java
```diff
@@ -25,6 +25,10 @@ CompositeTrackSelector create(Options options, DefaultBandwidthMeter bandwidthMe
                 Clock.DEFAULT
         );
         DefaultTrackSelector trackSelector = new DefaultTrackSelector(adaptiveTrackSelectionFactory);
+        DefaultTrackSelector.Parameters trackSelectorParameters = trackSelector.buildUponParameters()
+                .setMaxVideoBitrate(options.maxVideoBitrate())
```
frapontillo: We initialise the `DefaultTrackSelector` with the known `maxVideoBitrate` from the `options`.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/ExoPlayerTrackSelector.java
```diff
@@ -80,4 +80,20 @@ boolean supportsTrackSwitching(TrackType trackType,
...
+    void clearMaxVideoBitrate() {
+        setMaxVideoBitrateParameter(Integer.MAX_VALUE);
+    }
+
+    void setMaxVideoBitrate(int maxVideoBitrate) {
+        setMaxVideoBitrateParameter(maxVideoBitrate);
+    }
+
+    private void setMaxVideoBitrateParameter(int maxValue) {
+        trackSelector.setParameters(
```
frapontillo: This method also invalidates the player, so we do not need to reload anything.

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -322,4 +322,14 @@ float getVolume() {
         assertIsInPlaybackState();
         return volume;
     }
+
+    void clearMaxVideoBitrate() {
+        assertIsInPlaybackState();
+        NoPlayerLog.w("Tried to clear max video bitrate but has not been implemented for MediaPlayer.");
```
frapontillo: 🤷‍♂️ nothing we can do about it, I used the same way of no-opping we used in other methods in `AndroidMediaPlayerFacade`.

frapontillo:
ok to test

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
Mecharyry: It's going to be deprecated soon enough.

PR #214, opened 2019-04-08
NO-ADS/AdvertView

Mecharyry:
## Problem
We need to be able to setup and interact with an advert UI. 

## Solution
Exo-player provides an `AdsViewProvider` which is passed to the `AdsMediaSource` which eventually ends up in the load of the adverts in the `AdsLoader`. Unfortunately this interface deals with `ViewGroup`, which means we would end up casting/doing lots of manual work to make adverts work for our clients. 

I also found that the `AdsLoader.onAdClicked` doesn't go anywhere, there are no details of how you are expected to hook this up with the `AdViewProvider`.

In the end I opted for creating an `AdvertView` interface which can be bound to some client UI. It contains a `setup(List<AdvertBreak>, AdvertInteractionListener)` for dealing with creating the advert markers and the interaction with the adverts. I've added a preliminarily `removeMarker` for a future story but this is subject to change. 

I've created a `NoOp` version of the `AdvertView` because it is not essentially for the playback of adverts. I've also created a small wrapper around any actual `AdvertView` that is passed through to the loader that is `main thread` aware. This should forward all `UI` events from our background thread to the `UI` thread, ensuring that clients do not have any issues. 

### Test(s) added 
Yes, added some tests to the `AdsLoader`. I'd like to spend a little bit more time writing some more robust tests but these will come later. At the moment quite a lot of the events that we would want to test for are dependent on exo-player calls which we would have to duplicate.

## Follow-up
Add click through Urls to our advert models. 

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/AdvertView.java
+++ b/core/src/main/java/com/novoda/noplayer/AdvertView.java
@@ -1,0 +1,5 @@
+package com.novoda.noplayer;
+
+public interface AdvertView {
+
+}
--- a/core/src/main/java/com/novoda/noplayer/AdvertView.java
+++ b/core/src/main/java/com/novoda/noplayer/AdvertView.java
@@ -4,13 +4,11 @@ import java.util.List;
 
 public interface AdvertView {
 
-    void setup(List<AdvertBreak> advertBreaks);
+    void setup(List<AdvertBreak> advertBreaks, AdvertInteractionListener advertInteractionListener);
 
     void removeMarker(AdvertBreak advertBreak);
 
-    AdvertClickedListener getAdvertClickedListener();
-
-    interface AdvertClickedListener {
+    interface AdvertInteractionListener {
         void onAdvertClicked();
     }
 }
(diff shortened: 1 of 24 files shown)


core/src/main/java/com/novoda/noplayer/AdvertView.java
```diff
@@ -0,0 +1,14 @@
+package com.novoda.noplayer;
+
+import java.util.List;
+
+public interface AdvertView {
+
+    void setup(List<AdvertBreak> advertBreaks, AdvertInteractionListener advertInteractionListener);
```
Mecharyry: @JozefCeluch, maybe it is better to expose a `getAdvertListener` and hook that up to the other listeners. That way it all becomes push based. We could add another call on the advert listener like `onAdvertsLoaded(List<AdvertBreak>)`. 

Just thinking that `removeMarker` etc is oddly specific, might be better to keep it general.  What do you think?

Mecharyry: After thinking about it more I was a little worried about the threading if we took this approach but it seems that all listeners are called on the thread that the player is created with. In this case it would be the main thread.

PR #218, opened 2019-04-09
NO-ADS/Advert View

Mecharyry:
## Problem
When opening #214 I realised that the PR did not really match how the rest of `no-player` operates in terms of listeners and views. 

## Solution
Create an `AdvertView` that behaves similar to `PlayerView` in that it holds its own `AdvertListener`, which contains the generic advert events that occur. When an event occurs the UI can update accordingly. This `AdvertView` is passed to the `no-player` where the listener is bound to the collection of `AdvertListeners`. The `adsLoader`, which responds to user interactions, is bound to the `AdvertView`.

### Test(s) added 
Just updated some of the `AdsLoader` tests to ensure that the new `onAdvertsLoaded` is called once we have loaded the adverts.

### Paired with 
Nobody.


core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoOpAdvertListener.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.novoda.noplayer.Advert;
+import com.novoda.noplayer.AdvertBreak;
+import com.novoda.noplayer.AdvertBreakId;
+import com.novoda.noplayer.AdvertId;
+import com.novoda.noplayer.NoPlayer;
+
+import java.util.List;
+
+enum NoOpAdvertListener implements NoPlayer.AdvertListener {
```
Mecharyry: Moved this out of the `AdsLoader` to make it a little less HUGE

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -272,6 +273,18 @@ public void detach(PlayerView playerView) {
         this.playerView = null;
     }
 
+    @Override
+    public void attach(AdvertView advertView) {
+        listenersHolder.addAdvertListener(advertView.getAdvertListener());
```
Mecharyry: Attached to the collection of advert listeners so that the `AdvertView` and whatever other listeners the client attaches are notified of advert events.

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
```diff
@@ -334,4 +336,14 @@ void setMaxVideoBitrate(int maxVideoBitrate) {
         assertIsInPlaybackState();
         NoPlayerLog.w("Tried to set max video bitrate but has not been implemented for MediaPlayer.");
     }
+
+    void attach(AdvertView advertView) {
+        assertIsInPlaybackState();
+        NoPlayerLog.w("Tried to attach advert view but has not been implemented for MediaPlayer.");
```
Mecharyry: No point doing this for `media player` we will be deprecating it soon enough.

core/src/main/java/com/novoda/noplayer/AdvertView.java
```diff
@@ -0,0 +1,14 @@
...
+    NoPlayer.AdvertListener getAdvertListener();
+
+    void attach(AdvertInteractionListener advertInteractionListener);
+
+    void detach(AdvertInteractionListener advertInteractionListener);
+
+    interface AdvertInteractionListener {
+        void onAdvertClicked();
+    }
+}
```
pablisco: Should we have comments on this type and it's method? Given that it's a public API

core/src/main/java/com/novoda/noplayer/internal/mediaplayer/AndroidMediaPlayerFacade.java
pablisco: Would `assertIsInPlaybackState` crash if we aren't in the right state? What would happen to the log then?

Mecharyry:
It would swallow the log @pablisco but tbh if you are in the wrong state you have a bigger problem 😄

core/src/main/java/com/novoda/noplayer/AdvertView.java
Mecharyry: Absolutely, I'll add some now!

Mecharyry: Actually, would you be ok if I create a ticket for this and do a complete sweep of the library? Some other places, not related to this have been missed.

Mecharyry: https://github.com/novoda/no-player/issues/219

PR #221, opened 2019-04-10
NO-ADS/Skipping adverts bug

Mecharyry:
## Problem
It seems that when skipping to a resume point in the playback that occurs just after an advert break the advert start event is being emitted, even though we aren't starting an advert!

## Solution
First change `SkippedAdverts` to `PlayedAdverts`, we aren't really skipping them, we are resuming so they have been played before. 

Interestingly when the `onTimelineChanged` is called for this scenario the advert index that is reported may not have any adverts that are capable of being played, or intend to be played. In this case, we should not send the advert start event.

### Test(s) added 
Just updated the `PlayedAdverts` tests to represent `Played` instead of `Skipped`.

##### Beginning

| Before |
before_beginning |
| ------ | ----- |
| After| after_beginning |

##### Just after first advert break

Before | before_resuming_after_advert
--- | ---
After | after_resuming_after_advert

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
@@ -171,12 +171,19 @@ public class NoPlayerAdsLoader implements AdsLoader, Player.EventListener, Adver
         if (advertHasNotStarted()) {
             adGroupIndex = player.getCurrentAdGroupIndex();
             adIndexInGroup = player.getCurrentAdIndexInAdGroup();
-            notifyAdvertStart(advertBreaks.get(adGroupIndex));
+
+            if (hasUnplayedAdverts(adGroupIndex)) {
+                notifyAdvertStart(advertBreaks.get(adGroupIndex));
+            }
         }
     }
 
     private boolean advertHasNotStarted() {
         return player.isPlayingAd() && (adGroupIndex == -1 || adIndexInGroup == -1);
+    }
+
+    private boolean hasUnplayedAdverts(int adGroupIndex) {
+        return isPlayingAdvert() && adPlaybackState.adGroups[adGroupIndex].hasUnplayedAds();
     }
 
     private void notifyAdvertStart(AdvertBreak advertBreak) {
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
@@ -162,7 +162,7 @@ public class NoPlayerAdsLoader implements AdsLoader, Player.EventListener, Adver
         if (reason == Player.TIMELINE_CHANGE_REASON_PREPARED) {
             long contentPosition = player.getContentPosition();
             if (contentPosition > 0) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/PlayedAdverts.java
```diff
@@ -5,21 +5,25 @@
...
-        for (int i = advertBreaks.size() - 1; i >= 0; i--) {
-            AdvertBreak advertBreak = advertBreaks.get(i);
+        int numberOfAdvertBreaks = advertBreaks.size() - 1;
+        for (int advertBreakIndex = numberOfAdvertBreaks; advertBreakIndex >= 0; advertBreakIndex--) {
+            AdvertBreak advertBreak = advertBreaks.get(advertBreakIndex);
             if (advertBreak.startTimeInMillis() >= currentPositionInMillis) {
                 continue;
             }
 
-            adPlaybackStateWithSkippedAdGroups = adPlaybackStateWithSkippedAdGroups.withSkippedAdGroup(i);
```
JozefCeluch: Why can't we skip the whole group but we need to mark each advert in the group as played instead?

Mecharyry: Yes we can actually. I'll revert the change.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -171,14 +171,23 @@ public void onTimelineChanged(Timeline timeline, @Nullable Object manifest, int
         if (advertHasNotStarted()) {
             adGroupIndex = player.getCurrentAdGroupIndex();
             adIndexInGroup = player.getCurrentAdIndexInAdGroup();
-            notifyAdvertStart(advertBreaks.get(adGroupIndex));
+
+            if (canPlayAdverts(adGroupIndex)) {
```
Mecharyry: This checks the position information is correct within the `exo-player`, if it is and that particular advert has `AVAILABLE` adverts then we can play.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -205,7 +214,12 @@ public void onPositionDiscontinuity(int reason) {
...
```
Mecharyry: This should be the same as the `onTimelineChanged` code, it operates in the same way. We are always pairing the `start` and `end` events.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -171,14 +171,23 @@ public void onTimelineChanged(Timeline timeline, @Nullable Object manifest, int
...
+                notifyAdvertStart(advertBreaks.get(adGroupIndex));
+            } else {
+                resetAdvertPosition();
```
Mecharyry: Need to do this otherwise when you get to the next advert, after skipping one, the position data will still be present. It will then try to mark the skipped advert as `Played` which leads to an `IllegalStateException`

PR #222, opened 2019-04-11
NO-ADS/Enable Disable Adverts

Mecharyry:
## Problem
We want clients to be able to enable and disable adverts. Unfortunately this is not so straight forward with exo-player and it's `AdPlaybackState`. Let's go through the whole story in the Solution.

## Solution
First added the relevant checkbox etc into the UI to trigger the enabling and disabling of adverts. This calls through to the `NoPlayer` which calls down to the `NoPlayerAdvertLoader`.

I assumed that it would be as simple as replacing the current `AdPlaybackState` with an empty version and then when enabling replacing with the cached version. This works to a certain extent. Unfortunately adverts are gateways, this means that if you skip past an advert, either forwards or backwards, it will trigger the advert within that seek if it hasn't played yet. 

**What I opted to do in the end was:**

When disabling: 
- Skip all adverts that have not already played
- Create a flag that will prevent advert modifications when adverts are disabled, including seeks

When enabling: 
- Transform all `Skipped` adverts back to `Available` adverts if the advert start time is after the current player position. We keep all adverts before this time as skipped to prevent seek events going backwards from triggering the adverts. 

When seeking:
- If the advert start time is before the current position then it remains as Skipped, otherwise we transform it to Available and allow it to play. 

There is an `AvailableAdverts` transformer that relies on mutation...yes, mutation. We have to rely on mutation because `AdPlaybackState` does not allow adverts to transition from `Skipped` to `Available` in the public API. The only reasoning I have seen behind this decision is that the `AdPlaybackState` was created around a very specific usecase, `ImaAdsLoader`. 

I don't believe that this switching between `Skipped` and `Available` has any adverse affects on how adverts work, it seems like the `AdsMediaSource` does not do anything special when the `AdPlaybackState` is initially created, just initialises some maps with the correct sizes and suitable defaults. The `AdPlaybackState` is passed down to a `timeline` and a `media queue`, this timeline and media queue never loses information about these adverts, even if an advert subsequently becomes `Skipped` or `Unavailable`. 

Further reading -> https://github.com/google/ExoPlayer/issues/4005

### Test(s) added 
I've added tests to all the new classes that have been introduced that touch the `AdPlaybackState` in someway. 

### Paired with 
Nobody.


core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
```diff
@@ -0,0 +1,41 @@
...
+     * WARNING: This relies on mutation to achieve what would not be possible
+     * as the exo-player {@link AdPlaybackState} does not allow transitioning from
+     * Skipped to Available adverts.
+     *
+     * @param currentPositionInMillis The position after which all adverts will transition from Skipped to Available.
+     * @param advertBreaks            The client representation of the adverts, our source of truth.
+     * @param adPlaybackState         The {@link AdPlaybackState} to mutate with the new states.
+     * @return The {@link AdPlaybackState} with the new available states.
+     */
+    static AdPlaybackState from(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
```
Dorvaryn: I think we should not return anything and have a clearer method name to highlight what happens.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -203,19 +204,28 @@ private void notifyAdvertStart(AdvertBreak advertBreak) {
...
             // We need all of the above to be able to respond to advert events.
             return;
         }
 
-        if (isPlayingAdvert()) {
-            notifyAdvertEnd(advertBreaks.get(adGroupIndex));
-            adPlaybackState = adPlaybackState.withPlayedAd(adGroupIndex, adIndexInGroup);
+        if (reason == Player.DISCONTINUITY_REASON_SEEK) {
+            long contentPosition = player.getContentPosition();
+            adPlaybackState = AvailableAdverts.from(contentPosition, advertBreaks, this.adPlaybackState);
```
Dorvaryn: Why do we need to do this here ? This seems potentially taxing

core/src/main/java/com/novoda/noplayer/internal/exoplayer/PlayedAdverts.java
```diff
@@ -0,0 +1,39 @@
...
+     * Transforms all adverts before the current player position to the Played state.
+     * This is used to represent a resume position, in which a user would already seen the
+     * preceding adverts.
+     *
+     * @param currentPositionInMillis The position before which all adverts will transition from their current state to Played.
+     * @param advertBreaks            The client representation of the adverts, our source of truth.
+     * @param adPlaybackState         The {@link AdPlaybackState} to alter advert state on.
+     * @return The {@link AdPlaybackState} with the newly played states.
+     */
+    static AdPlaybackState from(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
```
Dorvaryn: Maybe change the method name for clarity

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SkippedAdverts.java
```diff
@@ -11,14 +11,16 @@ private SkippedAdverts() {
...
 
-    static AdPlaybackState from(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
+    /**
+     * Transforms all adverts that are not currently Played to Skipped.
+     *
+     * @param advertBreaks    The client representation of the adverts, our source of truth.
+     * @param adPlaybackState The {@link AdPlaybackState} to mutate with the new states.
+     * @return The {@link AdPlaybackState} with the new Skipped states.
+     */
+    static AdPlaybackState from(List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
```
Dorvaryn: Maybe change the method name for clarity

Mecharyry: 2d9062e

core/src/main/java/com/novoda/noplayer/internal/exoplayer/PlayedAdverts.java
Mecharyry: 9762514

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SkippedAdverts.java
Mecharyry: 18575bd

core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
Mecharyry: a1a8c63

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: We spoke about this offline. This is needed because when seeking backwards we need to be able to mark an advert as available to play again, that way when they seek forwards or the player naturally hits that time it will trigger the advert.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -343,4 +343,18 @@ void detach(AdvertView advertView) {
             advertView.detach(adsLoader);
         }
     }
+
+    void disableAdverts() {
+        if (adsLoader.isPresent()) {
+            NoPlayerAdsLoader adsLoader = this.adsLoader.get();
```
zegnus: is `this` needed?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -343,4 +343,18 @@ void detach(AdvertView advertView) {
...
+            adsLoader.disableAdverts();
+        }
+    }
+
+    void enableAdverts() {
+        if (adsLoader.isPresent()) {
+            NoPlayerAdsLoader adsLoader = this.adsLoader.get();
```
zegnus: same here

core/src/main/java/com/novoda/noplayer/internal/exoplayer/PlayedAdverts.java
```diff
@@ -0,0 +1,39 @@
...
+     */
+    static AdPlaybackState markAllPastAdvertsAsPlayed(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
+        AdPlaybackState adPlaybackStateWithPlayedAdGroups = adPlaybackState;
+        for (int advertBreakIndex = advertBreaks.size() - 1; advertBreakIndex >= 0; advertBreakIndex--) {
+            AdvertBreak advertBreak = advertBreaks.get(advertBreakIndex);
+            if (advertBreak.startTimeInMillis() >= currentPositionInMillis) {
+                continue;
+            }
+
+            for (int advertIndex = 0; advertIndex < advertBreak.adverts().size(); advertIndex++) {
```
zegnus: `advertBreak.adverts().size()` is a constant in this loop, we can extract it

core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
```diff
@@ -0,0 +1,38 @@
...
+     * @param currentPositionInMillis The position after which all adverts will transition from Skipped to Available.
+     * @param advertBreaks            The client representation of the adverts, our source of truth.
+     * @param adPlaybackState         The {@link AdPlaybackState} to mutate with the new states.
+     */
+    static void markAllFutureAdvertsAsAvailable(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
+        for (int i = advertBreaks.size() - 1; i >= 0; i--) {
+            AdvertBreak advertBreak = advertBreaks.get(i);
+            AdPlaybackState.AdGroup adGroup = adPlaybackState.adGroups[i];
+            if (advertBreak.startTimeInMillis() >= currentPositionInMillis) {
+                for (int stateIndex = 0; stateIndex < adGroup.states.length; stateIndex++) {
```
zegnus: `adGroup.states.length` is a constant in this loop, can be extracted

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -203,19 +204,28 @@ private void notifyAdvertStart(AdvertBreak advertBreak) {
...
             // We need all of the above to be able to respond to advert events.
             return;
         }
 
-        if (isPlayingAdvert()) {
-            notifyAdvertEnd(advertBreaks.get(adGroupIndex));
-            adPlaybackState = adPlaybackState.withPlayedAd(adGroupIndex, adIndexInGroup);
+        if (reason == Player.DISCONTINUITY_REASON_SEEK) {
+            long contentPosition = player.getContentPosition();
+            AvailableAdverts.markAllFutureAdvertsAsAvailable(contentPosition, advertBreaks, adPlaybackState);
```
zegnus: I need more info on this, calling you on slack

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
Mecharyry: yes, because the `adsLoader` on the left is a local variable and the `this` on the right is a field.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
Mecharyry: Same as above.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
zegnus: oh I see, be careful on that, it might be better to just rename the variables so that there is no need for `.this`

PR #223, opened 2019-04-12
NO-ADS/Add callback for played ads

JozefCeluch:
## Problem

We weren't notifying about ad breaks that were skipped due to playback starting at a different position other than 0.

## Solution

This adds a callback that's called after the `AdPlaybackState` is updated with the played ads. This just expands on the existing logic that was there already in the `PlayedAdverts` class and uses the same approach we used for `AdvertPlaybackState`.

I wasn't sure whether to call the method `onAdvertsSkipped` or `onAdvertsPlayed`. Alternatively we could also have a method `onAdvertBreakPlayed` (singular) that could be called for both when advert break ends and for each of the ones that are marked as played at the beginning.

### Test(s) added 

updated tests for `PlayedAdverts`

### Screenshots

| Before | After |
| ------ | ----- |
| gif/png _before_ | gif/png _after_ |

### Paired with 

nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/PlayedAdverts.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/PlayedAdverts.java
@@ -20,9 +20,9 @@ class PlayedAdverts {
      * @param currentPositionInMillis The position before which all adverts will transition from their current state to Played.
      * @param advertBreaks            The client representation of the adverts, our source of truth.
      * @param adPlaybackState         The {@link AdPlaybackState} to alter advert state on.
-     * @return The {@link AdPlaybackState} with the newly played states.
+     * @return The {@link PlayedAdverts} instance with the newly played states.
      */
-    static PlayedAdverts markAllPastAdvertsAsPlayed(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/PlayedAdverts.java
```diff
@@ -19,21 +20,36 @@ private PlayedAdverts() {
      * @param currentPositionInMillis The position before which all adverts will transition from their current state to Played.
      * @param advertBreaks            The client representation of the adverts, our source of truth.
      * @param adPlaybackState         The {@link AdPlaybackState} to alter advert state on.
-     * @return The {@link AdPlaybackState} with the newly played states.
+     * @return The {@link PlayedAdverts} instance with the newly played states.
      */
-    static AdPlaybackState markAllPastAdvertsAsPlayed(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
+    static PlayedAdverts from(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
```
Mecharyry: Can we go back to having the explicit name? `markAllPastAdvertsAsPlayed`? It still does the same from the looks of things 😄

JozefCeluch: We discussed this offline (via slack, so it was technically online), now this is just a standard static factory method that creates an instance of `PlayedAdverts` so the name is updated to match other factory methods in the project.

PR #230, opened 2019-04-29
Update snapshot releases

mr-archano:
## Problem

We are trying to migrate all our open source libraries to be hosted on a new Bintray account. The first step is to migrate the release script for the snapshots in order to use: https://bintray.com/novoda-oss/snapshots
A previous attempt at doing this (https://github.com/novoda/no-player/commit/1f22ae6a4647164466f1f236914123dfdb06e293) left the release of snapshots broken.

## Solution

Updated the release configuration to receive an additional `bintrayOrg`as build property. This value is provided to the CI job via a secret file containing all the Bintray-related config parameters, as per description in the [build script](https://github.com/novoda/no-player/blob/a8c0915c7e8d1785b46cd4ea2dd114909aa3997a/core/build.gradle#L18-L22):
```
- bintrayOrg: name of the Bintray organisation to deploy the artifacts to
- bintrayRepo: name of the repo of the organisation to deploy the artifacts to
- bintrayUser: name of the account used to deploy the artifacts
- bintrayKey: API key of the account used to deploy the artifacts
```

As part of the migration we had also to use a different naming pattern for snapshot builds, as the new account doesn't support artifact versions with `SNAPSHOT` in them.

### Test(s) added 

No tests added. Updated README to point to new snapshot repo. Once this is merged we should see a new snapshot linked in the README.

### Paired with 

Nobody

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,19 +79,19 @@ compileOptions {
 
 ## Snapshots
 
-[![CI status](https://ci.novoda.com/buildStatus/icon?job=no-player-snapshot)](https://ci.novoda.com/job/no-player-snapshot/lastBuild/console) [![Download from Bintray](https://api.bintray.com/packages/novoda/snapshots/no-player/images/download.svg)](https://bintray.com/novoda/snapshots/no-player/_latestVersion)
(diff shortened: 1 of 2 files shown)


mr-archano:
@Mecharyry if you don't want to migrate the snapshots to use the new Bintray account yet then we could revert the changes to the README and configure the CI job to use the old account credentials (I've re-uploaded the secret files missing the `bintrayOrg` value)

PR #231, opened 2019-04-29
Update Gradle to latest release

mr-archano:
## Problem
After some attempts at using the latest Gradle and Android Gradle Plugin version we had to downgrade because of lack of support in https://github.com/novoda/bintray-release.

## Solution

The `bintray-release` plugin has been patched to support Gradle 5.2-5.4.1 for Android projects, so we should be able to use Gradle 5.4.1 and Android Gradle Plugin 3.4.0 without issues.

### Paired with 

Nobody


build.gradle
```diff
@@ -11,11 +11,14 @@ buildscript {
     repositories {
         google()
         jcenter()
+        maven {
+            url 'https://dl.bintray.com/novoda-oss/snapshots'
```
mr-archano: Note: unfortunately we are having some issues with `bintray-release`, as the latest version of the plugin `0.9.1`is not showing up in JCenter (the issue might be related to the migration of Bintray accounts that we are performing).
In the meantime we have decided to use a snapshot release  of the plugin instead.

PR #232, opened 2019-04-29
Notify Security Level

Mecharyry:
## Problem
We want to be able to see what DRM security level we are going to be using when creating a `MediaDrm` session. 

## Solution
Create a `DrmSecurityLevelCallback` which will notify clients after a DRM session is created of the security level in use for the content. 

### Test(s) added 
Just updated where necessary. 

### Paired with 
Nobody.


Mecharyry:
I'm doing some additional changes to force a security level on initial creation of a DRM session.

PR #235, opened 2019-05-07
Advert Error prevents content playing

Mecharyry:
## Problem
In the event that we receive an error when loading adverts we want to continue playback rather than failing altogether. This issue occurs because the `AdsMediaSource` will not load anything until an `AdPlaybackState` is passed to it. 

## Solution
When an error occurs during loading of adverts emit the error to `NoPlayer.AdvertListener.onAdvertsLoadError(Exception)` and update the `AdPlaybackState`, the latter of which will force content to begin playback. 

In the event that a prepare error occurs, this is when the advert is preparing for playback, forward the advert and any associated exception to `NoPlayer.AdvertListener.onAdvertPrepareError(Advert, IOException)`.

**But Ryan couldn't you just use the `AdsLoader.EventListener` from `exo-player`?** 
Yes I could. This would emit to the rather more obscure `MediaSourceEventListener.onLoadError` from which the client or us, at a different layer, would need to look through the information and decide what to do with it. You get info like: `MediaLoadData`, `IOException`, `wasCanceled` etc. 

### Test(s) added 
Updated the tests for the `NoPlayerAdsLoader` to ensure that the `AdvertListener` is called for error events.

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/AdvertsLoader.java
+++ b/core/src/main/java/com/novoda/noplayer/AdvertsLoader.java
@@ -16,7 +16,7 @@ public interface AdvertsLoader {
     interface Callback {
         void onAdvertsLoaded(List<AdvertBreak> advertBreaks);
 
-        void onAdvertsError(String message);
+        void onAdvertsError(Exception cause);
     }
 
     interface Cancellable {
--- a/core/src/main/java/com/novoda/noplayer/NoPlayer.java
+++ b/core/src/main/java/com/novoda/noplayer/NoPlayer.java
@@ -312,7 +312,7 @@ public interface NoPlayer extends PlayerState {
 
     interface AdvertListener {
 
-        void onAdvertsLoadError(String message);
+        void onAdvertsLoadError(Exception cause);
 
         void onAdvertsLoaded(List<AdvertBreak> advertBreaks);
 
--- a/core/src/main/java/com/novoda/noplayer/NoPlayer.java
+++ b/core/src/main/java/com/novoda/noplayer/NoPlayer.java
@@ -11,6 +11,7 @@ import com.novoda.noplayer.model.PlayerSubtitleTrack;
 import com.novoda.noplayer.model.PlayerVideoTrack;
 import com.novoda.noplayer.model.Timeout;
 
+import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 
@@ -311,11 +312,15 @@ public interface NoPlayer extends PlayerState {
 
     interface AdvertListener {
 
+        void onAdvertsLoadError(String message);
+
         void onAdvertsLoaded(List<AdvertBreak> advertBreaks);
 
         void onAdvertBreakStart(AdvertBreak advertBreak);
 
         void onAdvertBreakEnd(AdvertBreak advertBreak);
+
+        void onAdvertPrepareError(Advert advert, IOException failedWith);
 
         void onAdvertStart(Advert advert);
 
@@ -328,6 +333,7 @@ public interface NoPlayer extends PlayerState {
         void onAdvertsEnabled(List<AdvertBreak> advertBreaks);
 
         void onAdvertsSkipped(List<AdvertBreak> advertBreaks);
+
     }
 
     /**
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
@@ -94,14 +94,14 @@ public class NoPlayerAdsLoader implements AdsLoader, Player.EventListener, Adver
 
         @Override
         public void onAdvertsError(final Exception cause) {
+            loadingAds = null;
+            advertBreaks = Collections.emptyList();
+            adPlaybackState = new AdPlaybackState();
+
             handler.post(new Runnable() {
                 @Override
                 public void run() {
-                    loadingAds = null;
-
-                    adPlaybackState = new AdPlaybackState();
                     updateAdPlaybackState();
-
                     advertListener.onAdvertsLoadError(cause);
                 }
             });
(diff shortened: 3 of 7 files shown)


core/src/main/java/com/novoda/noplayer/AdvertsLoader.java
```diff
@@ -16,7 +16,7 @@
     interface Callback {
         void onAdvertsLoaded(List<AdvertBreak> advertBreaks);
 
-        void onAdvertsError(String message);
+        void onAdvertsError(Exception cause);
```
Mecharyry: Made more sense in the end not to limit the information for an error to a `String`, will help our clients diagnose issues if they occur.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -93,9 +93,18 @@ public void run() {
...
 
         @Override
-        public void onAdvertsError(String message) {
+        public void onAdvertsError(final Exception cause) {
             loadingAds = null;
-            eventListener.onAdLoadError(null, null);
+            advertBreaks = Collections.emptyList();
+            adPlaybackState = new AdPlaybackState();
+
+            handler.post(new Runnable() {
```
Mecharyry: All callbacks should be performed on the `main-thread` as it is the one from which the player is created.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -150,6 +159,10 @@ public void release() {
     public void handlePrepareError(int adGroupIndex, int adIndexInAdGroup, IOException exception) {
         if (adPlaybackState != null) {
             adPlaybackState = adPlaybackState.withAdLoadError(adGroupIndex, adIndexInAdGroup);
+
+            AdvertBreak advertBreak = advertBreaks.get(adGroupIndex);
+            Advert advert = advertBreak.adverts().get(adIndexInAdGroup);
+            advertListener.onAdvertPrepareError(advert, exception);
```
Mecharyry: Pass back as much information as possible.

Mecharyry: Should never be passed `-1` or incorrect values as it is directly correlated to the `AdPlaybackState` which in turn is created from our models.

Mecharyry: We receive all of these on the `main-thread` so we don't need to worry about using a handler here.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -150,6 +159,11 @@ public void release() {
     public void handlePrepareError(int adGroupIndex, int adIndexInAdGroup, IOException exception) {
         if (adPlaybackState != null) {
             adPlaybackState = adPlaybackState.withAdLoadError(adGroupIndex, adIndexInAdGroup);
+            updateAdPlaybackState();
```
Mecharyry: Yea we actually need this 😂 if we don't do this then the `AdPlaybackState` is marked with an error but is never passed to the underlying `source` and so playback just stops. No advert events, nothing 😂

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
zegnus: I guess that we assume that those gets are not going to return nulls or throw exceptions...

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -93,9 +93,18 @@ public void run() {
...
+        public void onAdvertsError(final Exception cause) {
             loadingAds = null;
-            eventListener.onAdLoadError(null, null);
+            advertBreaks = Collections.emptyList();
+            adPlaybackState = new AdPlaybackState();
+
+            handler.post(new Runnable() {
+                @Override
+                public void run() {
+                    updateAdPlaybackState();
```
Mecharyry: `AdsMediaSource` doesn't load anything until a `AdPlaybackState` is passed to it. Even if adverts fail we need to pass this source in order for the content to play.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: They can't return `null`, if it contained a `null` for an `advertBreak` then it would fail in another place when loading the adverts. As for throwing, it shouldn't throw, these indexes are defined by the source which comes from our playback state, which again, comes from our models.

core/src/main/java/com/novoda/noplayer/NoPlayer.java
```diff
@@ -311,12 +312,16 @@
 
     interface AdvertListener {
 
+        void onAdvertsLoadError(Exception cause);
```
JozefCeluch: Does this notify about the same exception we pass in to the `AdvertsLoader.Callback`?

core/src/main/java/com/novoda/noplayer/NoPlayer.java
```diff
@@ -311,12 +312,16 @@
...
+
         void onAdvertsLoaded(List<AdvertBreak> advertBreaks);
 
         void onAdvertBreakStart(AdvertBreak advertBreak);
 
         void onAdvertBreakEnd(AdvertBreak advertBreak);
 
+        void onAdvertPrepareError(Advert advert, IOException failedWith);
```
JozefCeluch: Why not just use `cause` instead of `failedWith`?

Also... how about adding some javadoc? 😝

core/src/main/java/com/novoda/noplayer/NoPlayer.java
JozefCeluch: yes is the answer, I just needed to scroll down 😂

PR #239, opened 2019-05-13
Update dependencies

zegnus:
## Problem

Update dependencies

## Solution

- ExoPlayer updated to 2.9.10 https://github.com/google/ExoPlayer/blob/release-v2/RELEASENOTES.md#2100
- Project auto-migrated to AndroidX

### Test(s) added 

All tested and video playback tested through the demo app

### Screenshots

No UI changes

### Paired with 

Nobody


core/src/test/java/com/novoda/noplayer/internal/exoplayer/SecurityDowngradingCodecSelectorTest.java
```diff
@@ -30,21 +31,21 @@
     public void whenContentIsSecure_thenRequiresSecureDecoderIsFalse() throws MediaCodecUtil.DecoderQueryException {
         SecurityDowngradingCodecSelector securityDowngradingCodecSelector = new SecurityDowngradingCodecSelector(internalMediaCodecUtil);
 
-        securityDowngradingCodecSelector.getDecoderInfos(ANY_MIME_TYPE, CONTENT_SECURE);
+        securityDowngradingCodecSelector.getDecoderInfos(ANY_MIME_TYPE, CONTENT_SECURE, DOES_NOT_REQUIRES_TUNNELING_DECODER);
```
Mecharyry:
```suggestion
        securityDowngradingCodecSelector.getDecoderInfos(ANY_MIME_TYPE, CONTENT_SECURE, DOES_NOT_REQUIRE_TUNNELING_DECODER);
```

core/src/test/java/com/novoda/noplayer/internal/exoplayer/SecurityDowngradingCodecSelectorTest.java
```diff
@@ -19,6 +19,7 @@
 
     private static final boolean CONTENT_SECURE = true;
     private static final boolean CONTENT_INSECURE = false;
+    private static final boolean DOES_NOT_REQUIRES_TUNNELING_DECODER = false;
```
Mecharyry:
```suggestion
    private static final boolean DOES_NOT_REQUIRE_TUNNELING_DECODER = false;
```

core/src/test/java/com/novoda/noplayer/internal/exoplayer/SecurityDowngradingCodecSelectorTest.java
```diff
@@ -30,21 +31,21 @@
     public void whenContentIsSecure_thenRequiresSecureDecoderIsFalse() throws MediaCodecUtil.DecoderQueryException {
         SecurityDowngradingCodecSelector securityDowngradingCodecSelector = new SecurityDowngradingCodecSelector(internalMediaCodecUtil);
 
-        securityDowngradingCodecSelector.getDecoderInfos(ANY_MIME_TYPE, CONTENT_SECURE);
+        securityDowngradingCodecSelector.getDecoderInfos(ANY_MIME_TYPE, CONTENT_SECURE, DOES_NOT_REQUIRES_TUNNELING_DECODER);
 
         ArgumentCaptor<Boolean> argumentCaptor = ArgumentCaptor.forClass(Boolean.class);
-        verify(internalMediaCodecUtil).getDecoderInfos(eq(ANY_MIME_TYPE), argumentCaptor.capture());
+        verify(internalMediaCodecUtil).getDecoderInfos(eq(ANY_MIME_TYPE), argumentCaptor.capture(), eq(DOES_NOT_REQUIRES_TUNNELING_DECODER));
```
Mecharyry:
```suggestion
        verify(internalMediaCodecUtil).getDecoderInfos(eq(ANY_MIME_TYPE), argumentCaptor.capture(), eq(DOES_NOT_REQUIRE_TUNNELING_DECODER));
```

core/src/test/java/com/novoda/noplayer/internal/exoplayer/SecurityDowngradingCodecSelectorTest.java
```diff
@@ -30,21 +31,21 @@
...
         assertThat(argumentCaptor.getValue()).isFalse();
     }
 
     @Test
     public void whenContentIsInsecure_thenRequiresSecureDecoderIsFalse() throws MediaCodecUtil.DecoderQueryException {
         SecurityDowngradingCodecSelector securityDowngradingCodecSelector = new SecurityDowngradingCodecSelector(internalMediaCodecUtil);
 
-        securityDowngradingCodecSelector.getDecoderInfos(ANY_MIME_TYPE, CONTENT_INSECURE);
+        securityDowngradingCodecSelector.getDecoderInfos(ANY_MIME_TYPE, CONTENT_INSECURE, DOES_NOT_REQUIRES_TUNNELING_DECODER);
```
Mecharyry:
```suggestion
        securityDowngradingCodecSelector.getDecoderInfos(ANY_MIME_TYPE, CONTENT_INSECURE, DOES_NOT_REQUIRE_TUNNELING_DECODER);
```

core/src/test/java/com/novoda/noplayer/internal/exoplayer/SecurityDowngradingCodecSelectorTest.java
```diff
@@ -30,21 +31,21 @@
...
 
         ArgumentCaptor<Boolean> argumentCaptor = ArgumentCaptor.forClass(Boolean.class);
-        verify(internalMediaCodecUtil).getDecoderInfos(eq(ANY_MIME_TYPE), argumentCaptor.capture());
+        verify(internalMediaCodecUtil).getDecoderInfos(eq(ANY_MIME_TYPE), argumentCaptor.capture(), eq(DOES_NOT_REQUIRES_TUNNELING_DECODER));
```
Mecharyry:
```suggestion
        verify(internalMediaCodecUtil).getDecoderInfos(eq(ANY_MIME_TYPE), argumentCaptor.capture(), eq(DOES_NOT_REQUIRE_TUNNELING_DECODER));
```

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SimpleRenderersFactory.java
```diff
@@ -165,6 +167,7 @@ private void buildVideoRenderers(Context context,
                 allowedVideoJoiningTimeMs,
                 drmSessionManager,
                 DO_NOT_PLAY_CLEAR_SAMPLES_WITHOUT_KEYS,
+                ENABLE_DECODER_FALLBACK,
```
Mecharyry: > This may result in using a decoder that is slower/less efficient than the primary decoder.

I'm wondering if we should be attempting to log stuff like this when it happens. Slower or less efficient may lead to problems during playback.

JozefCeluch: Isn't this a replacement for `MediaCodecSelector.DEFAULT_WITH_FALLBACK`? We provide a flag to use that on the builder so this should be in sync with that value

Mecharyry: Yes. My point was more about the effect that this could have in general regardless of how it was implemented. 

If we use a fallback decoder then we could potentially have playback issues like stuttering, because we are using a less efficient decoder. It was a thought, just thinking out loud 😄

JozefCeluch: Yes, that should be a separate issue that we can look into regardless of this change.

My concern was more about the fact that this seems to be changing the behavior in an unexpected way since it will always enable the fallback regardless of the value that the clients use in the builder.

Mecharyry: I guess you are referring to https://github.com/novoda/no-player/blob/master/core/src/main/java/com/novoda/noplayer/PlayerBuilder.java#L102? This is not the same thing 😄 the one in the builder refers to force downgrading to use an insecure decoder when we are able to. The one changed in this PR is going to use a fallback decoder when we fail to initialise the one we want.

PR #240, opened 2019-05-16
Notify Track Changes

Mecharyry:
## Problem
It's possible that the tracks for audio, video and subtitles may not be ready by the time we are playing content! This is most apparent when we are switching between adverts, we will receive notifications for adverts ending and starting before the tracks themselves become available!

## Solution
Add a new callback that notifies the client when `tracks` have been changed. Clients can then use this to show and hide any track selector UI they may have. 

### Screenshots

| Audio Only | Video Only |
| ------ | ----- |
audio_only | asset_video_only

## Improvements for future
Ideally from the exo-player:

```
default void onTracksChanged(
        TrackGroupArray trackGroups, TrackSelectionArray trackSelections) {}
```

we would map to a generic `no-player` Track, unfortunately all of our tracks at the moment are dependent on the type of format they are representing i.e. Subtitles vs Video. The first step would be to make these generic and then emit this from this call. 

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/listeners/TracksChangedListeners.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/listeners/TracksChangedListeners.java
@@ -1,0 +1,30 @@
+package com.novoda.noplayer.internal.listeners;
+
+import com.novoda.noplayer.NoPlayer;
+
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+class TracksChangedListeners implements NoPlayer.TracksChangedListener {
+
+    private final Set<NoPlayer.TracksChangedListener> listeners = new CopyOnWriteArraySet<>();
+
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/novoda/noplayer/internal/listeners/TracksChangedListeners.java
```diff
@@ -0,0 +1,30 @@
+package com.novoda.noplayer.internal.listeners;
+
+import com.novoda.noplayer.NoPlayer;
+
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+class TracksChangedListeners implements NoPlayer.TracksChangedListener {
+
+    private final Set<NoPlayer.TracksChangedListener> listeners = new CopyOnWriteArraySet<>();
```
zegnus: 👍

PR #241, opened 2019-05-17
Set advert break initial position

JozefCeluch:
## Problem

There was no way to set the initial position of an advert break so when the video would be reloaded from the position of the advert break, the advert break would always start from beginning.

## Solution

This adds `initialAdvertBreakPositionInMillis` to `Options` that are passed in when the video is loaded. This API matches the `PlayerState.positionInAdvertBreakInMillis()` method. 

Internally we then iterate over the adverts inside the first advert break and mark them as played if their duration is smaller than the progress in advert break. Once we reach the advert that would finish after the provided position, we set the remaining time to be the `adResumePositionUs` in `AdPlaybackState`.

At the moment the marking of adverts as played happens on the first group. I was wondering if it would make sense to tie this marking to the `initialPositionInMillis` and only modify the advert break which start time is the same as the initial position. That way we would be able to ensure that the progress inside an advert break is only modified if the advert break is the first thing that's being played when the content starts.

### Test(s) added 

added tests for the creation of `AdvertPlaybackState` and modified and added where necessary to match the change in `Options`. I also create `AdGroupFixture` to make it easier to create `AdGroup`s in tests.

### Paired with 

nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackState.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackState.java
@@ -14,14 +14,11 @@ final class AdvertPlaybackState {
     private final AdPlaybackState adPlaybackState;
     private final List<AdvertBreak> advertBreaks;
 
-    static AdvertPlaybackState from(List<AdvertBreak> advertBreaks, long advertBreakResumePositionMillis) {
-        AdvertPlaybackState advertPlaybackState = from(advertBreaks);
-        AdPlaybackState adPlaybackState = advertPlaybackState.adPlaybackState;
-        adPlaybackState = updateResumePositionInFirstGroup(adPlaybackState, advertBreakResumePositionMillis);
-        return new AdvertPlaybackState(adPlaybackState, advertPlaybackState.advertBreaks);
+    static AdvertPlaybackState from(List<AdvertBreak> advertBreaks) {
+        return from(advertBreaks, 0);
     }
 
-    static AdvertPlaybackState from(List<AdvertBreak> advertBreaks) {
+    static AdvertPlaybackState from(List<AdvertBreak> advertBreaks, long advertBreakResumePositionMillis) {
         List<AdvertBreak> sortedAdvertBreaks = sortAdvertBreaksByStartTime(advertBreaks);
 
         long[] advertOffsets = advertBreakOffset(sortedAdvertBreaks);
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackState.java
```diff
@@ -42,9 +46,37 @@ static AdvertPlaybackState from(List<AdvertBreak> advertBreaks) {
         }
 
         adPlaybackState = adPlaybackState.withAdDurationsUs(advertBreaksWithAdvertDurations);
+        adPlaybackState = updateResumePositionInFirstGroup(adPlaybackState, advertBreakResumePositionMillis);
         return new AdvertPlaybackState(adPlaybackState, sortedAdvertBreaks);
     }
 
+    private static AdPlaybackState updateResumePositionInFirstGroup(AdPlaybackState state, long positionMillis) {
```
danybony: Would it make sense to move this into `AdPlaybackState`, like `withAdDurationsUs()` so that we have `adPlaybackState.withResumePositionInFirstGroup()`?

JozefCeluch: `AdPlaybackState` is an ExoPlayer class, that's why we have all this logic in a similar-sounding `AdvertPlaybackState` 😄

core/src/test/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackStateTest.java
```diff
@@ -47,19 +50,88 @@
...
+                .withPlayedStateAt(0)
+                .withPlayedStateAt(1)
+                .build();
+        long resumePosition = FIRST_ADVERT.durationInMillis() + SECOND_ADVERT.durationInMillis() + HALF_SECOND_IN_MILLIS;
+
+        AdvertPlaybackState advertPlaybackState = AdvertPlaybackState.from(advertBreaks, resumePosition);
+        AdPlaybackState adPlaybackState = advertPlaybackState.adPlaybackState();
+
+        assertThat(adPlaybackState.adGroups).containsExactly(expectedAdGroup);
+        assertThat(adPlaybackState.adResumePositionUs).isEqualTo(HALF_SECOND_IN_MICROS);
```
juankysoriano: We should be able to assert here that `adPlaybackState` `STATE` is `AD_STATE_PLAYED`; right? 

And that will honor even more the name of the test

JozefCeluch: I guess that's part of the weirdness of the `AdPlaybackState`. At the top of the test lines `99` and `100` we create the expected group with the played adverts at the expected positions. But then the `adResumePositionUs` is not per group but it's global in the whole `AdPlaybackState` so that's why we only assert it here.

PR #244, opened 2019-05-30
Video type is lost when error is notified

JozefCeluch:
## Problem

Error listener was notified after the player was released which meant that we were not able to decide what type of video the player was playing when the error happened.

## Solution

To fix this I register the listener that resets the player to be called _after_ all the client listeners are called. This is done in the `ExoPlayerForwarder` which binds the internal error listener after the error listeners from the `PlayerListenerHolder`.

I also changed the API to check the `VideoType` that is set to be played. Before we had two separate boolean methods but that makes them a bit hard to work with since the advert and content are mutually exclusive (ie. player can only play either content or advert and not both). But the player can also play nothing if it's not prepared. Having all these states in single method makes is easier to understand all the possible cases.

### Test(s) added 

updated where needed for the new behavior

### Screenshots

no UI changes

### Paired with 

nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
@@ -71,14 +71,6 @@ class ExoPlayerFacade {
         return exoPlayer != null && exoPlayer.getPlayWhenReady();
     }
 
-    boolean isSetToPlayAdvert() {
-        return exoPlayer != null && exoPlayer.isPlayingAd();
-    }
-
-    boolean isSetToPlayContent() {
-        return exoPlayer != null && !isSetToPlayAdvert();
-    }
-
     PlayerState.VideoType videoType() {
         if (exoPlayer == null) {
             return PlayerState.VideoType.UNDEFINED;
@@ -131,6 +123,11 @@ class ExoPlayerFacade {
 
         return 0;
     }
+
+    private boolean isSetToPlayAdvert() {
+        return videoType() == PlayerState.VideoType.ADVERT;
+    }
+
 
     private long combinedAdvertDurationInGroup(Timeline.Period period, int numberOfAdvertsToInclude) {
         int adGroupIndex = exoPlayer.getCurrentAdGroupIndex();
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
@@ -1,9 +1,7 @@
 package com.novoda.noplayer.internal.exoplayer;
 
 import android.net.Uri;
-
 import androidx.annotation.Nullable;
-
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.Player;
 import com.google.android.exoplayer2.SimpleExoPlayer;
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerForwarder.java
```diff
@@ -58,8 +58,9 @@ public void bind(NoPlayer.CompletionListener completionListener, NoPlayer.StateC
         exoPlayerEventListener.add(new OnCompletionStateChangedForwarder(stateChangedListener));
     }
 
-    public void bind(NoPlayer.ErrorListener errorListener) {
+    public void bind(NoPlayer.ErrorListener errorListener, NoPlayer.ErrorListener internalErrorListener) {
         exoPlayerEventListener.add(new PlayerOnErrorForwarder(errorListener));
+        exoPlayerEventListener.add(new PlayerOnErrorForwarder(internalErrorListener));
```
JozefCeluch: the `internalErrorListener` will be called after `errorListener` because `ExoPlayerEventListener` adds all the listeners to a `List`

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerForwarder.java
```diff
@@ -58,8 +58,9 @@ public void bind(NoPlayer.CompletionListener completionListener, NoPlayer.StateC
         exoPlayerEventListener.add(new OnCompletionStateChangedForwarder(stateChangedListener));
     }
 
-    public void bind(NoPlayer.ErrorListener errorListener) {
+    public void bind(NoPlayer.ErrorListener errorListener, NoPlayer.ErrorListener internalErrorListener) {
```
zegnus: worth collapsin these params into a list `...`? as public method, you can add a javadoc explaining that the order of the listeners matter

JozefCeluch: That's exactly why I didn't use `varargs` 😄. But maybe I shouldn't even be modifying this and just call the method twice and test that it's called in the right order

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -72,12 +71,14 @@ boolean isPlaying() {
         return exoPlayer != null && exoPlayer.getPlayWhenReady();
     }
 
-    boolean isSetToPlayAdvert() {
-        return exoPlayer != null && exoPlayer.isPlayingAd();
-    }
-
-    boolean isSetToPlayContent() {
-        return exoPlayer != null && !isSetToPlayAdvert();
+    PlayerState.VideoType videoType() {
```
juankysoriano: :100:

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
```diff
@@ -123,6 +124,11 @@ long positionInAdvertBreakInMillis() throws IllegalStateException {
         return 0;
     }
 
+    private boolean isSetToPlayAdvert() {
```
juankysoriano: Do we still need this one? I think we can now simply inline it?, anyway, not a blocker, it's a private method so...

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerTwoImpl.java
```diff
@@ -74,18 +72,13 @@ void initialise() {
         forwarder.bind(listenersHolder.getDroppedVideoFramesListeners());
         forwarder.bind(listenersHolder.getAdvertListeners());
         forwarder.bind(listenersHolder.getTracksChangedListeners());
+        forwarder.bind(resetOnErrorListener());
```
juankysoriano: :+1:

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ExoPlayerFacade.java
JozefCeluch: I kept it because it's used twice, breaking the triple-usage rule 😄 YOLO

PR #245, opened 2019-05-30
Mark adverts as skipped before content resume position

JozefCeluch:
## Problem

When a content is started with a resume position then the player always plays the advert that is before the initial position (same issue as #243).

## Solution

The fix is the same as in (#243), that is using the initial content position in the same way as seek position was used before. For that we need to pass the initial position to the `NoPlayerAdsLoader`.

I also fixed a bug where only the earliest advert block would be marked as skipped.

### Test(s) added 

updated where necessary to include the new changes, and I also modified the existing test to reveal the bug

### Screenshots

n/a

### Paired with 

nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/SkippedAdverts.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/SkippedAdverts.java
@@ -43,15 +43,15 @@ final class SkippedAdverts {
     static AdPlaybackState markAllPastAvailableAdvertsAsSkipped(long currentPositionInMillis,
                                                                 List<AdvertBreak> advertBreaks,
                                                                 AdPlaybackState adPlaybackState) {
-        AdPlaybackState adPlaybackStateWithSkippedAdGroups = adPlaybackState;
+        AdPlaybackState updatedPlaybackState = adPlaybackState;
         for (int i = advertBreaks.size() - 1; i >= 0; i--) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/SkippedAdverts.java
```diff
@@ -43,15 +43,15 @@ static AdPlaybackState markAllNonPlayedAdvertsAsSkipped(List<AdvertBreak> advert
     static AdPlaybackState markAllPastAvailableAdvertsAsSkipped(long currentPositionInMillis,
                                                                 List<AdvertBreak> advertBreaks,
                                                                 AdPlaybackState adPlaybackState) {
-        AdPlaybackState adPlaybackStateWithSkippedAdGroups = adPlaybackState;
+        AdPlaybackState updatedPlaybackState = adPlaybackState;
         for (int i = advertBreaks.size() - 1; i >= 0; i--) {
             AdvertBreak advertBreak = advertBreaks.get(i);
-            AdPlaybackState.AdGroup adGroup = adPlaybackState.adGroups[i];
```
JozefCeluch: here we were always modifying the initial state so we would only mark the last accessed group as skipped

JozefCeluch:
@zegnus any tests would require a lot setup and mocking so it wouldn't be actually testing much in the end. Maybe there's a room to extract some collaborators from there, but that would need to be done separately

PR #246, opened 2019-06-05
Allow no-player to skip a given advert break

juankysoriano:
## Problem

It is in some case convenient to be able to skip just the current advert break, or the current advert. 

## Solution

Allow `NoPlayer` to `skipAdvert` or `skipAdvertBreak`

### Test(s) added 

Yes, to make sure only the given `AdvertBreak`, or `Advert` is skipped

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/NoPlayer.java
+++ b/core/src/main/java/com/novoda/noplayer/NoPlayer.java
@@ -136,7 +136,7 @@ public interface NoPlayer extends PlayerState {
 
     void disableAdverts();
 
-    void skipAdverts(@NonNull AdvertBreak advertBreak);
+    void skipAdvertBreak(@NonNull AdvertBreak advertBreak);
 
     void enableAdverts();
 
--- a/core/src/main/java/com/novoda/noplayer/NoPlayer.java
+++ b/core/src/main/java/com/novoda/noplayer/NoPlayer.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Map;
 
 import androidx.annotation.FloatRange;
+import androidx.annotation.NonNull;
 
 // There are a lot of features for playing and monitoring video.
 @SuppressWarnings("PMD.ExcessivePublicCount")
@@ -134,6 +135,8 @@ public interface NoPlayer extends PlayerState {
     void detach(AdvertView advertView);
 
     void disableAdverts();
+
+    void skipAdverts(@NonNull AdvertBreak advertBreak);
 
     void enableAdverts();
 
@@ -393,6 +396,12 @@ public interface NoPlayer extends PlayerState {
          */
         void onAdvertsEnabled(List<AdvertBreak> advertBreaks);
 
+        /**
+         * Called when advert break is skipped or prior to being skipped
+         *
+         * @param advertBreak that is being skipped.
+         */
+        void onAdvertsSkipped(AdvertBreak advertBreak);
     }
 
     /**
--- a/core/src/main/java/com/novoda/noplayer/NoPlayer.java
+++ b/core/src/main/java/com/novoda/noplayer/NoPlayer.java
@@ -16,7 +16,6 @@ import java.util.List;
 import java.util.Map;
 
 import androidx.annotation.FloatRange;
-import androidx.annotation.NonNull;
 
 // There are a lot of features for playing and monitoring video.
 @SuppressWarnings("PMD.ExcessivePublicCount")
@@ -136,7 +135,9 @@ public interface NoPlayer extends PlayerState {
 
     void disableAdverts();
 
-    void skipAdvertBreak(@NonNull AdvertBreak advertBreak);
+    void skipAdvertBreak();
+
+    void skipAdvert();
 
     void enableAdverts();
 
@@ -401,7 +402,14 @@ public interface NoPlayer extends PlayerState {
          *
          * @param advertBreak that is being skipped.
          */
-        void onAdvertsSkipped(AdvertBreak advertBreak);
+        void onAdvertBreakSkipped(AdvertBreak advertBreak);
+
+        /**
+         * Called when advert is skipped or prior to being skipped
+         *
+         * @param advert that is being skipped.
+         */
+        void onAdvertSkipped(Advert advert);
     }
 
     /**
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
@@ -35,4 +35,16 @@ final class AvailableAdverts {
             }
         }
     }
+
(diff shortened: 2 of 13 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
```diff
@@ -35,4 +35,16 @@ static void markAllFutureAdvertsAsAvailable(long currentPositionInMillis, List<A
             }
         }
     }
+
+    static void markAllFutureAdvertsAsPlayed(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
```
JozefCeluch: this method should probably be removed as it seems to be unused

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SkippedAdverts.java
```diff
@@ -32,6 +32,21 @@ static AdPlaybackState markAllNonPlayedAdvertsAsSkipped(List<AdvertBreak> advert
...
+     * @param advertBreaks    The client representation of all the adverts break.
+     * @param advertBreak    The client representation of the advert break to be skipped.
+     * @param adPlaybackState The {@link AdPlaybackState} to alter advert state on.
+     * @return The {@link AdPlaybackState} with the new Skipped states.
+     */
+    static AdPlaybackState markAllNonPlayedAdvertsAsSkipped(AdvertBreak advertBreak, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
+        AdPlaybackState adPlaybackStateWithSkippedAdGroups = adPlaybackState;
+        int breakIndex = advertBreaks.indexOf(advertBreak);
+        adPlaybackStateWithSkippedAdGroups = adPlaybackStateWithSkippedAdGroups.withSkippedAdGroup(breakIndex);
+        return adPlaybackStateWithSkippedAdGroups;
```
JozefCeluch: seems like we could just return the result on the previous line and we wouldn't need the extra `adPlaybackStateWithSkippedAdGroups` variable

core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
Mecharyry: This is in the wrong class. Or at least, the name of this class now needs to change. The adverts wouldn't be available if they are marked as `Played`.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SkippedAdverts.java
```diff
@@ -32,6 +32,21 @@ static AdPlaybackState markAllNonPlayedAdvertsAsSkipped(List<AdvertBreak> advert
...
 
+    /**
+     * Transforms all adverts in the given break that are not currently Played to Skipped.
+     *
+     * @param advertBreaks    The client representation of all the adverts break.
+     * @param advertBreak    The client representation of the advert break to be skipped.
+     * @param adPlaybackState The {@link AdPlaybackState} to alter advert state on.
+     * @return The {@link AdPlaybackState} with the new Skipped states.
+     */
+    static AdPlaybackState markAllNonPlayedAdvertsAsSkipped(AdvertBreak advertBreak, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
```
JozefCeluch: this method should probably be called something else to be more descriptive, maybe `markAdvertBreakAsSkipped`?

core/src/main/java/com/novoda/noplayer/NoPlayer.java
```diff
@@ -135,6 +136,8 @@
 
     void disableAdverts();
 
+    void skipAdverts(@NonNull AdvertBreak advertBreak);
```
Mecharyry: I'd probably drop the `@NonNull` from all of these, I find this to be a redundant annotation. The implication should always be to pass something that is `@NonNull`

JozefCeluch: this is not related to this PR, but we should consider adding `@NonNull` annotation to the public API because otherwise Kotlin thinks that all the objects there are nullable when it's not the case

core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
Mecharyry: Ah wait. This isn't used...

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -293,6 +296,17 @@ void disableAdverts() {
         advertsDisabled = true;
     }
 
+    void skipAdverts(@NonNull AdvertBreak advertBreak) {
```
Mecharyry: I think this just happens to work because the `disable` and `enable` aren't called again in the same playback session. I would expect this to not work if the disable and then the enable were called immediately. I think this is fine, we are kind of backed into a corner as far as the states go.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SkippedAdverts.java
```diff
@@ -32,6 +32,21 @@ static AdPlaybackState markAllNonPlayedAdvertsAsSkipped(List<AdvertBreak> advert
         return adPlaybackStateWithSkippedAdGroups;
     }
 
+    /**
+     * Transforms all adverts in the given break that are not currently Played to Skipped.
+     *
+     * @param advertBreaks    The client representation of all the adverts break.
+     * @param advertBreak    The client representation of the advert break to be skipped.
```
Mecharyry: I think we need to run the auto-formatter on this class, this alignment looks a bit out.

core/src/main/java/com/novoda/noplayer/NoPlayer.java
Mecharyry: Makes sense 👍

core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
juankysoriano: 4fd5fddb

core/src/main/java/com/novoda/noplayer/NoPlayer.java
juankysoriano: I'll keep the `@NonNull` on public ones

core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
juankysoriano: I started doing it, but never used :man_facepalming:

PR #247, opened 2019-06-05
Ignore track changes when seek and track groups are empty

JozefCeluch:
## Problem

ExoPlayer notifies about incorrect track changes after seeking when adverts were disabled.

The issue is that if we disable adverts and then seek, we get 2 `onTracksChanged()` notifications. The first one is with empty tracks, and only the second one contains the tracks we had before. If we propagate that to the client they might incorrectly assume that there is an error with the tracks they should be seeing.

## Solution
Save the reason passed into `onPositionDiscontinuity()` and use it to understand what if we should filter out the `onTracksChanged()` notification. We also need to check for empty tracks because otherwise we wouldn't propagate the `onTracksChanged()` when the video starts.

### Test(s) added 

yes, for the modified forwarder

### Paired with 

nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/TracksChangedForwarder.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/TracksChangedForwarder.java
@@ -11,6 +11,7 @@ import com.novoda.noplayer.NoPlayer;
 class TracksChangedForwarder implements Player.EventListener {
 
     private final NoPlayer.TracksChangedListener tracksChangedListener;
+    private int discontinuityReason = -1;
 
     TracksChangedForwarder(NoPlayer.TracksChangedListener tracksChangedListener) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/TracksChangedForwarder.java
```diff
@@ -38,7 +39,11 @@ public void onTimelineChanged(Timeline timeline, Object manifest, @Player.Timeli
 
     @Override
     public void onTracksChanged(TrackGroupArray trackGroups, TrackSelectionArray trackSelections) {
-        tracksChangedListener.onTracksChanged();
+        if (discontinuityReason == Player.DISCONTINUITY_REASON_SEEK && trackGroups.isEmpty()) {
```
juankysoriano: :+1:

PR #248, opened 2019-06-05
Notify advert break ends if last advert if skipped

juankysoriano:
## Problem

When the last advert of a break is skipped we are not notifying that the advert break ended

## Solution

When skipping an advert, if the advert is the last one on the break, then notify advert break end

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
@@ -313,8 +313,14 @@ public class NoPlayerAdsLoader implements AdsLoader, Player.EventListener, Adver
 
         adPlaybackState = SkippedAdverts.markAdvertAsSkipped(adIndexInGroup, adGroupIndex, adPlaybackState);
         updateAdPlaybackState();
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -313,7 +313,12 @@ void skipAdvert() {
 
         adPlaybackState = SkippedAdverts.markAdvertAsSkipped(adIndexInGroup, adGroupIndex, adPlaybackState);
         updateAdPlaybackState();
-        advertListener.onAdvertSkipped(advertBreaks.get(adGroupIndex).adverts().get(adIndexInGroup));
+        AdvertBreak advertBreak = advertBreaks.get(adGroupIndex);
+        List<Advert> adverts = advertBreak.adverts();
+        advertListener.onAdvertSkipped(adverts.get(adIndexInGroup));
+        if (adIndexInGroup == adverts.size() - 1) {
+            advertListener.onAdvertBreakEnd(advertBreak);
```
Mecharyry: I'm ok with this workaround for now. We should rewrite the `onDiscontinuity` and `timeline` methods later to make this more robust.

PR #250, opened 2019-06-06
Manually fallback insecure decoder

zegnus:
## Problem

There is a performance issue with Huawei P20 and other devices.

## Solution

Huawei P20 requires a secure decoder while Intel devices do not support secure decoders. In order to solve this for those two scenarios, when we require a downgrade of the decoder, we are going to try to request if the device supports a secure decoder if it this we will return a secure one. If the device does not have any, then we will request an unsecured list of decoders.

### Test(s) added 

Manually tested in different devices

### Screenshots

No UI changes

### Paired with 

Nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
+++ b/core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
@@ -107,8 +107,7 @@ public class PlayerBuilder {
     }
 
     /**
-     * Forces secure decoder selection to be ignored in favour of using an insecure decoder.
-     * e.g. Forcing an L3 stream to play with an L3 decoder instead of an L1 secure decoder by default.
+     * Allows the selection of a insecure decoder when the device does not support a secure decoder.
      *
      * @return {@link PlayerBuilder}
      */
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
```diff
@@ -107,8 +107,7 @@ public PlayerBuilder withPriority(PlayerType playerType, PlayerType... playerTyp
     }
 
     /**
-     * Forces secure decoder selection to be ignored in favour of using an insecure decoder.
-     * e.g. Forcing an L3 stream to play with an L3 decoder instead of an L1 secure decoder by default.
+     * Allows the selection of a insecure decoder when the device does not support a secure decoder.
```
Mecharyry:
```suggestion
     * Allows the selection of an insecure decoder when the device does not support a secure decoder.
```

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SecurityDowngradingCodecSelector.java
```diff
@@ -6,9 +6,14 @@
 
 import java.util.List;
 
+/**
+ * This class will try and get a decoder that requires secure decryption and fallback
+ * to a decoder that does not require secure decryption if there is non available
```
Mecharyry:
```suggestion
 * to a decoder that does not require secure decryption if there is non available.
```

Mecharyry:
```suggestion
 * to a decoder that does not require secure decryption if there is none available.
```

PR #251, opened 2019-06-07
Show adverts even on seek

Mecharyry:
## Problem
We have been attempting to stop adverts from playing when a user performs a seek but isn't passing through an advert marker. Instead we are now going to allow the content to be gated. Users have to watch an advert in order to watch a section of content. 

## Solution
Mainly removing code around the seek. Also made it so that skipped adverts are all re-enabled when enabling adverts, rather than just those in the future. 

### Test(s) added 
Updated where necessary.

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
@@ -18,19 +18,16 @@ final class AvailableAdverts {
      * as the exo-player {@link AdPlaybackState} does not allow transitioning from
      * Skipped to Available adverts.
      *
-     * @param currentPositionInMillis The position after which all adverts will transition from Skipped to Available.
-     * @param advertBreaks            The client representation of the adverts, our source of truth.
-     * @param adPlaybackState         The {@link AdPlaybackState} to mutate with the new states.
+     * @param advertBreaks    The client representation of the adverts, our source of truth.
+     * @param adPlaybackState The {@link AdPlaybackState} to mutate with the new states.
      */
-    static void markAllFutureAdvertsAsAvailable(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
+    static void markSkippedAdvertsAsAvailable(List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
         for (int i = advertBreaks.size() - 1; i >= 0; i--) {
             AdvertBreak advertBreak = advertBreaks.get(i);
             AdPlaybackState.AdGroup adGroup = adPlaybackState.adGroups[i];
-            if (advertBreak.startTimeInMillis() >= currentPositionInMillis) {
-                for (int stateIndex = 0; stateIndex < adGroup.states.length; stateIndex++) {
-                    if (adGroup.states[stateIndex] == AdPlaybackState.AD_STATE_SKIPPED) {
-                        adGroup.states[stateIndex] = AdPlaybackState.AD_STATE_AVAILABLE;
-                    }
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/AvailableAdverts.java
```diff
@@ -18,19 +18,16 @@ private AvailableAdverts() {
...
      * Skipped to Available adverts.
      *
-     * @param currentPositionInMillis The position after which all adverts will transition from Skipped to Available.
-     * @param advertBreaks            The client representation of the adverts, our source of truth.
-     * @param adPlaybackState         The {@link AdPlaybackState} to mutate with the new states.
+     * @param advertBreaks    The client representation of the adverts, our source of truth.
+     * @param adPlaybackState The {@link AdPlaybackState} to mutate with the new states.
      */
-    static void markAllFutureAdvertsAsAvailable(long currentPositionInMillis, List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
+    static void markSkippedAdvertsAsAvailable(List<AdvertBreak> advertBreaks, AdPlaybackState adPlaybackState) {
```
Mecharyry: We no longer make adverts only in the future available, we make all adverts available again. The only way users cannot see an advert again is to watch it all the way to the end so that it is marked as `PLAYED`.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -95,9 +89,7 @@ public void onAdvertsLoaded(List<AdvertBreak> breaks) {
             AdvertPlaybackState advertPlaybackState = AdvertPlaybackState.from(breaks, advertBreakResumePosition);
             advertBreaks = advertPlaybackState.advertBreaks();
             adPlaybackState = advertPlaybackState.adPlaybackState();
-            if (contentInitialPosition > 0) {
```
Mecharyry: We've had a few chats about this now and if the user triggers playback from a resume position then they are subject to any advert that may be associated with that section of content.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -327,26 +303,10 @@ void enableAdverts() {
...
-
-    @Override
-    public void onBeat(NoPlayer noPlayer) {
-        if (player == null || player.isPlayingAd()) {
-            return;
-        }
-
-        lastContentPositionInMillis = player.getContentPosition();
-
-        if (hasPerformedSeek) {
```
Mecharyry: We aren't concerned with disabling adverts on seek now, they can play based on the state they are currently in.

core/src/test/java/com/novoda/noplayer/internal/exoplayer/AvailableAdvertsTest.java
```diff
@@ -45,7 +45,7 @@
     @Test
     public void makesSkippedAdvertsAfterCurrentOrAtCurrentPositionAvailable() {
         AdPlaybackState adPlaybackState = adPlaybackStateWithSkippedAdverts();
-        AvailableAdverts.markAllFutureAdvertsAsAvailable(THIRTY_SECONDS_IN_MILLIS, ADVERT_BREAKS, adPlaybackState);
```
Mecharyry: Wait, these aren't correct. I'll update this.

Mecharyry: 0aedeac

PR #252, opened 2019-06-07
Allow secure and non-secure decoders

Mecharyry:
## Problem
Tried the changes in #250 on a client application, my pixel stopped playing content 😭 

## Solution
Turns out that we should probably try to use a secure decoder first and then switch to the non-secure version if it fails. 

### Paired with 
Nobody.


core/src/main/java/com/novoda/noplayer/internal/exoplayer/SecurityDowngradingCodecSelector.java
```diff
@@ -32,19 +33,20 @@ public static SecurityDowngradingCodecSelector newInstance() {
             boolean requiresSecureDecoder,
             boolean requiresTunnelingDecoder
     ) throws MediaCodecUtil.DecoderQueryException {
-        List<MediaCodecInfo> decoderInfos = internalMediaCodecUtil.getDecoderInfos(
+        List<MediaCodecInfo> decoderInfos = new ArrayList<>(internalMediaCodecUtil.getDecoderInfos(
```
Mecharyry: The initial collection is not modifiable.

PR #254, opened 2019-06-13
Secure and non-secure codecs

Mecharyry:
## Problem
We ended up having to revert the codec selection changes (#253) because we were attempting to grab secure and non-secure codecs for content that did not have a DRM requirement 😬 

## Solution
~~Create a `CodecSecurityRequirement` which is just a singleton around a boolean which represents whether we want secure codecs or not. The magic actually happens in our new `SecurityRequirementBasedMediaCodecVideoRenderer` which updates this boolean based on a `Format`s `DrmInit` data at the points before the `codecs` are read from the system.~~

I realised late in the day yesterday that I had overcomplicated everything 😂 so it turns out the `supportFormat` is used to determine whether or not a given renderer is capable of supporting a given format. This `supportFormat` will use the other `renderer` method `getDecoderInfos` for determining the base codecs it will use. In the event that this `getDecoderInfos` returns an empty list it will attempt to change the parameters slightly based on `DrmInitData` etc and return a slightly different support level based on it. Given we only have a single renderer for `Video`, as we do not enable extensions, this method is completely redundant pretty much. 

This means we can just look at the `Renderer.getDecoderInfos`. This one actually has the `Format` which means it knows about the `DrmInitData` that we use to select decoders. This simplifies everything, A LOT. The solution is two parts. 

- One, the `CodecSelector` which will select secure codecs when `DrmInitData` is available.
- Two, a new `Renderer` which overrides the `requiresSecureDecoder` that we pass to the `CodecSelector`. This `Renderer` will just check for `DrmInitData`, it won't go further than that. If it's present then use `secure` followed by `non-secure`. If the `DrmInitData` is missing, just use the `non-secure`. 

I've updated the `PlayerBuilder` to make the Codec selection more obvious. We aren't downgrading every time, only in the instances where the initialisation fails.  

#### Why a singleton
The other option was to pass down a concrete type of a `MediaCodecSelector` or implement an additional interface. Felt it was better to keep the two parts separate. We have a singleton which is always updated but not always used, the `MediaCodecSelector` decides whether it needs to do anything with it. 

### Test(s) added 
Just updated where necessary.

### Paired with 
Nobody.


core/src/main/java/com/novoda/noplayer/internal/exoplayer/SecurityRequirementBasedMediaCodecVideoRenderer.java
```diff
@@ -0,0 +1,73 @@
...
+                playClearSamplesWithoutKeys,
+                enableDecoderFallback,
+                eventHandler,
+                eventListener,
+                maxDroppedFramesToNotify
+        );
+    }
+
+    @Override
+    protected int supportsFormat(MediaCodecSelector mediaCodecSelector,
```
Mecharyry: These two methods are the only usages I have seen of the `MediaCodecSelector`. I have not applied the same logic to the `AudioRenderers` so they will use whatever is dictated by this I guess.

Mecharyry: @zegnus do you think it is worth doing the same for the `audio renderer`?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SecurityRequirementBasedMediaCodecVideoRenderer.java
```diff
@@ -0,0 +1,73 @@
...
+                                 DrmSessionManager<FrameworkMediaCrypto> drmSessionManager,
+                                 Format format) throws MediaCodecUtil.DecoderQueryException {
+        CodecSecurityRequirement codecSecurityRequirement = CodecSecurityRequirement.getInstance();
+        DrmInitData drmInitData = format.drmInitData;
```
zegnus: so this is actually saying that if the content is not DRM, then don't bother trying secure decoders, yup?

Mecharyry: Yea, if the content doesn't use DRM then there is no point using a secure decoder. It doesn't require it after all 😄

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SecurityRequirementBasedMediaCodecVideoRenderer.java
zegnus: I think it is yes, now that we are on it let's get it sorted

juankysoriano:
The `SecureCodecsRequired` singleton get's updated in a couple of places as far as I can see.

The question is, is there any risk about for example `MediaVideoRenderer` updating it saying "this is required" while `MediaAudioRenderer` updating it afterwards saying "this is not required"? If this happens then we'll have an incorrect `boolean` on it

Mecharyry:
Good question @juankysoriano. This is why the `SecureCodecsRequirement` is updated each time before a read occurs, in that renderer, to ensure that the value is correct before a read occurs. 

The only other option I can think of is to create a new interface that allows us to change the state on the instance. We then pass a media codec selector for video and audio separately. Do you think that would be better? 🤔

core/src/main/java/com/novoda/noplayer/internal/exoplayer/MediaCodecAudioRendererWithSimplifiedDrmRequirement.java
```diff
@@ -0,0 +1,51 @@
...
+                audioCapabilities,
+                audioProcessors
+        );
+    }
+
+    @Override
+    protected List<MediaCodecInfo> getDecoderInfos(MediaCodecSelector mediaCodecSelector,
+                                                   Format format,
+                                                   boolean requiresSecureDecoder) throws MediaCodecUtil.DecoderQueryException {
+        return super.getDecoderInfos(mediaCodecSelector, format, format.drmInitData != null);
```
Mecharyry: Here we relax the rules that would normally determine the `requiresSecureDecoder`. Normally it would check the data inside the `DrmInitData`. We don't want to go that far, we are optimistic. If the secure one fails, we just use a `non-secure` version.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/MediaCodecVideoRendererWithSimplifiedDrmRequirement.java
```diff
@@ -0,0 +1,51 @@
...
+                eventListener,
+                maxDroppedFramesToNotify
+        );
+    }
+
+    @Override
+    protected List<MediaCodecInfo> getDecoderInfos(MediaCodecSelector mediaCodecSelector,
+                                                   Format format,
+                                                   boolean requiresSecureDecoder) throws MediaCodecUtil.DecoderQueryException {
+        return super.getDecoderInfos(mediaCodecSelector, format, format.drmInitData != null);
```
Mecharyry: Same as the `audio`, just for `video`. Relax those rules.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerExoPlayerCreator.java
```diff
@@ -87,10 +86,6 @@ ExoPlayerTwoImpl create(Context context,
                     allowCrossProtocolRedirects
             );
 
-            MediaCodecSelector mediaCodecSelector = downgradeSecureDecoder
```
Mecharyry: We pass the boolean down to where it is used now. It goes to a factory that handles the construction and usage of this.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SimpleRenderersFactory.java
```diff
@@ -162,15 +166,31 @@ private void buildVideoRenderers(Context context,
...
-                mediaCodecSelector,
-                allowedVideoJoiningTimeMs,
-                drmSessionManager,
-                DO_NOT_PLAY_CLEAR_SAMPLES_WITHOUT_KEYS,
-                ENABLE_DECODER_FALLBACK,
-                eventHandler,
-                eventListener,
-                MAX_DROPPED_VIDEO_FRAME_COUNT_TO_NOTIFY));
+        if (allowFallbackDecoder) {
+            outRenderers.add(new MediaCodecVideoRendererWithSimplifiedDrmRequirement(
```
Mecharyry: When we allow fallbacks: 

- We relax the DRM rules in the renderer so that it will require secure codecs for all `Format`s that have `DrmInitData` regardless of what it is. 
- We use the `MediaCodecSelector` that is capable of providing `secure` and `non-secure` codecs.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/SimpleRenderersFactory.java
```diff
@@ -218,18 +240,29 @@ private void buildAudioRenderers(Context context,
...
-                drmSessionManager,
-                PLAY_CLEAR_SAMPLES_WITHOUT_KEYS,
-                eventHandler,
-                eventListener,
-                AudioCapabilities.getCapabilities(context),
-                audioProcessors
-        );
-
-        outRenderers.add(mediaCodecAudioRenderer);
+        if (allowFallbackDecoder) {
```
Mecharyry: Same as above. 🏈

PR #255, opened 2019-06-14
Move retrieval of Key Set Id to Local DRM Session Manager

juankysoriano:
## Problem

The retrieval of the `Key Set Id` could be a long running task that at the moment we are performing upon `ExoPlayer` creation. But there is no need of doing it there.
## Solution

Move the retrieval of the `Key Set Id` via `DownloadedModularDrm` to `LocalSessionManager.acquireSession` which should be a thread safe place

### Test(s) added 

Yes, to verify that the `KeySetId` is retrieved correctly when required


juankysoriano:
I will double check that

Mecharyry:
> which should be a thread safe place

But do we know for sure that it is? Did we check strict-mode?

Mecharyry:
Screenshot 2019-06-18 at 09 17 17

^^ Thread that is used inside that `LocalDrmSessionManager`. I think we might be able to use the `DefaultDrmSessionManager` to perform what we want. I think we could even do this using the `StreamingModularDrm` as the `byte[]` is the raw `KeySetId`.

PR #256, opened 2019-06-17
Advert resume position

Mecharyry:
## Problem
We've seen adverts resuming from the beginning or at odd times when resuming playback 😬 Seems to occur when the first advert duration is greater than the resume position BUT the second advert is less than the resume position. 

**Resume: 2000**
**Advert 1: 3000**
**Advert 2: 1000**

Resume is less than 3000 so we don't update the `played advert duration` BUT we also don't break out because the check is against the `played advert duration` which we didn't update 😬 

So:

Resume is greater than 1000 so we do update the `played advert duration`

At which point we the do `resume - duration` which would be `2000 - 1000` which would not match the original position 😬 

## Solution
I've moved the `ResumeableAdverts` away from the `AdvertPlaybackState`, it felt like a separate operation that can be performed after the base creation has been performed. It also made testing a little easier because I only needed to be concerned with a single part. 

The actual solution is just ensuring that we perform one operation. Either we update the duration and continue through the adverts until out criteria is met OR our criteria is already met, in which case, we break out. At least one of these operations must occur each time we try to resume. 

### Test(s) added 
Yes, moved to a separate class and updated the tests to show the error case.

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackState.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackState.java
@@ -15,10 +15,6 @@ final class AdvertPlaybackState {
     private final List<AdvertBreak> advertBreaks;
 
     static AdvertPlaybackState from(List<AdvertBreak> advertBreaks) {
-        return from(advertBreaks, 0);
-    }
-
-    static AdvertPlaybackState from(List<AdvertBreak> advertBreaks, long advertBreakResumePositionMillis) {
         List<AdvertBreak> sortedAdvertBreaks = sortAdvertBreaksByStartTime(advertBreaks);
 
         long[] advertOffsets = advertBreakOffset(sortedAdvertBreaks);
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackState.java
```diff
@@ -46,37 +42,9 @@ static AdvertPlaybackState from(List<AdvertBreak> advertBreaks, long advertBreak
...
-
-        AdPlaybackState updatedState = state;
-        long playedAdvertDuration = 0;
-        for (int index = 0; index < firstAdGroup.count; index++) {
-            long durationWithCurrentAd = playedAdvertDuration + firstAdGroup.durationsUs[index];
-            if (durationWithCurrentAd <= groupResumePosition) {
-                updatedState = updatedState.withPlayedAd(0, index);
-                playedAdvertDuration += firstAdGroup.durationsUs[index];
-            }
-            if (groupResumePosition <= playedAdvertDuration) {
```
Mecharyry: This was the problem. The `playedAdvertDuration` is only updated when the resume position is greater than the adverts watched duration. However we don't break out if this first condition is met, so we can iterate through the adverts until we find one that matches our criteria. If we do this then the resume position updated on line 74 evaluates incorrectly.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/ResumeableAdverts.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.android.exoplayer2.C;
+import com.google.android.exoplayer2.source.ads.AdPlaybackState;
+
+final class ResumeableAdverts {
+
+    private ResumeableAdverts() {
+        // Utility class.
+    }
+
+    static AdPlaybackState markAsResumeableFrom(AdPlaybackState adPlaybackState, long positionInMillis) {
```
Mecharyry: Moved to a separate class because it felt like an additional operation **after** the initial creation, also helped with testing.

core/src/test/java/com/novoda/noplayer/internal/exoplayer/AdvertPlaybackStateTest.java
```diff
@@ -82,58 +84,6 @@ public void createsEmptyStateWhenAdvertBreaksAreEmpty() {
         assertThat(adPlaybackState.contentDurationUs).isEqualTo(C.TIME_UNSET);
     }
 
-    @Test
-    public void createsAdvertPlaybackStateWithResumePositionInMicroseconds() {
```
Mecharyry: Moved the tests related to the `resume position` to the new class.

core/src/test/java/com/novoda/noplayer/internal/exoplayer/ResumeableAdvertsTest.java
```diff
@@ -0,0 +1,125 @@
...
+            .build();
+    private static final AdvertBreak SECOND_ADVERT_BREAK = anAdvertBreak()
+            .withStartTimeInMillis(TWENTY_SECONDS_IN_MILLIS)
+            .withAdverts(THIRD_ADVERT)
+            .build();
+    private static final List<AdvertBreak> ADVERT_BREAKS = Arrays.asList(FIRST_ADVERT_BREAK, SECOND_ADVERT_BREAK);
+    private static final AdPlaybackState.AdGroup SECOND_AD_GROUP = secondAdGroupFixture().build();
+
+    @Test
+    public void createsAdvertPlaybackStateWithResumePositionInMicroseconds() {
```
Mecharyry: This is the test case that breaks. If we have the first advert having a duration greater than the resume position and then the second advert having a duration that is less than the resume position.

PR #257, opened 2019-06-19
Custom playback parameters

Mecharyry:
## Problem
Trying to create a new flow for downloaded and streaming DRM but it is very hard to work with the demo because we do not try to download a `KeySetId`. 

## Solution
Create a `LandingActivity` which allows users of the demo to specify a custom `mpd` address, `license` address and whether the `Drm` `KeySetId` should be downloaded. 

## Follow up
The next PR will add the `exo-player` to the demo sources and will use the `OfflineLicenseHelper` to download and store the `KetSetId` for subsequent playback. 

### Screenshots
custom_uri

### Paired with 
Nobody.


demo/build.gradle
```diff
@@ -24,6 +24,8 @@ android {
 }
 dependencies {
     implementation project(':core')
+    implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
+    implementation 'com.google.android.material:material:1.1.0-alpha07'
```
zegnus: using material for a demo ✨

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -94,20 +117,27 @@ public void onTracksChanged() {
...
         });
+    }
 
+    private void extractFromIntent() {
+        Intent intent = getIntent();
+        if (intent != null) {
+            mpdAddress = Uri.parse(intent.getStringExtra(LandingActivity.KEY_MPD_ADDRESS));
+            licenseServerAddress = intent.getStringExtra(LandingActivity.KEY_LICENSE_SERVER_ADDRESS);
+            downloadLicense = intent.getBooleanExtra(LandingActivity.KEY_DOWNLOAD_LICENSE, false);
+        }
```
zegnus: shall we show a toast indicating that there is an issue? so that the user of the demo knows what needs to be done

Mecharyry: Toast when any of these are missing? I can do.

Mecharyry: They'd all cancel each other though if each was missing. I think if it doesn't work it is quite obvious that information is missing from the `LandingActivity`.

Mecharyry: 5626621

PR #258, opened 2019-06-19
Downloaded license

Mecharyry:
## Problem
#257 introduced custom playback parameters via a `LandingActivity`, including the ability to specify whether the DRM KeySetId was downloaded. We should probably go ahead and download the `KeySetId` and use it in this case.

## Solution
Introduce an `OfflineLicense` which uses the `ExoPlayer` `OfflineLicenseHelper` under the hood to provision and request a DRM session, we then extract the relevant `KeySetId` and store for usage by the player. 

## Next Steps
Remove the `LocalDrmSessionManager` replacing it with the `DefaultDrmSessionManager`. The `Default` can perform all of the heavy lifting for us and probably has better error handling in the event something goes wrong.

### Paired with 
Nobody.

### DIFF
--- a/demo/src/main/java/com/novoda/demo/MainActivity.java
+++ b/demo/src/main/java/com/novoda/demo/MainActivity.java
@@ -4,21 +4,14 @@ import android.app.Activity;
 import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
-import android.os.Handler;
-import android.os.Looper;
 import android.util.Log;
 import android.view.View;
 import android.widget.CheckBox;
 import android.widget.CompoundButton;
 import android.widget.Toast;
 
-import com.google.android.exoplayer2.drm.DrmInitData;
-import com.google.android.exoplayer2.drm.DrmSession;
 import com.google.android.exoplayer2.drm.OfflineLicenseHelper;
 import com.google.android.exoplayer2.drm.UnsupportedDrmException;
-import com.google.android.exoplayer2.source.dash.DashUtil;
-import com.google.android.exoplayer2.source.dash.manifest.DashManifest;
-import com.google.android.exoplayer2.upstream.DataSource;
 import com.google.android.exoplayer2.upstream.DefaultHttpDataSourceFactory;
 import com.novoda.noplayer.ContentType;
 import com.novoda.noplayer.NoPlayer;
@@ -35,7 +28,6 @@ import com.novoda.noplayer.model.KeySetId;
 import com.novoda.noplayer.model.PlayerSubtitleTrack;
 import com.novoda.noplayer.model.PlayerVideoTrack;
 
-import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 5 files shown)


demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -68,7 +87,7 @@ protected void onCreate(Bundle savedInstanceState) {
             drmHandler = new DownloadedModularDrm() {
                 @Override
                 public KeySetId getKeySetId() {
-                    return null; //TODO: Actually download the keyset id.
+                    return KeySetId.of(offlineKeySetId);
```
danybony: This will crash with a NPE in case the license is not downloaded yet.
Should we fail gracefully and return null here instead?

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -142,13 +161,19 @@ private void extractFromIntent() {
...
     protected void onStart() {
         super.onStart();
-        Options options = new OptionsBuilder()
-                .withContentType(ContentType.DASH)
-                .withMinDurationBeforeQualityIncreaseInMillis(HALF_A_SECOND_IN_MILLIS)
-                .withMaxInitialBitrate(TWO_MEGABITS)
-                .withMaxVideoBitrate(getMaxVideoBitrate())
-                .build();
-        demoPresenter.startPresenting(mpdAddress, options);
+        offlineLicense.download(new OfflineLicense.OfflineLicenseCallback() {
```
danybony: NPE in case of `UnsupportedDrmException` when creating the `offlineLicense`. Should we care?

demo/src/main/java/com/novoda/demo/OfflineLicense.java
```diff
@@ -0,0 +1,78 @@
...
+                    });
+                } catch (DrmSession.DrmSessionException e) {
+                    Log.e(getClass().getSimpleName(), "DrmSession.DrmSessionException", e);
+                    Toast.makeText(context, "DrmSession.DrmSessionException: " + e.getMessage(), Toast.LENGTH_SHORT).show();
+                } catch (IOException e) {
+                    Log.e(getClass().getSimpleName(), "IOException", e);
+                    Toast.makeText(context, "IOException:" + e.getMessage(), Toast.LENGTH_SHORT).show();
+                } catch (InterruptedException e) {
+                    Log.e(getClass().getSimpleName(), "InterruptedException", e);
+                    Toast.makeText(context, "InterruptedException:" + e.getMessage(), Toast.LENGTH_SHORT).show();
```
danybony: Are we not using Java <insert here 7 or 8 I don't remember>?
We could collapse the three catch into one and make the messages depending on fields in the exception

demo/src/main/java/com/novoda/demo/MainActivity.java
Mecharyry: Very true 😄 I'll change this to return null.

demo/src/main/java/com/novoda/demo/MainActivity.java
Mecharyry: I think this would only be the case if the device didn't support widevine which is probably not the case of the people that use this library. Or at least, won't be for much longer considering we will drop `MediaPlayer`.

demo/src/main/java/com/novoda/demo/OfflineLicense.java
Mecharyry: Yea great point 😄 that will make it a little less messy.

demo/src/main/java/com/novoda/demo/MainActivity.java
Mecharyry: 75e2ad4

demo/src/main/java/com/novoda/demo/OfflineLicense.java
Mecharyry: 105dbab

PR #260, opened 2019-06-20
Advert loader state rework

Mecharyry:
⚠️ This is quite a big change, so we need to have a discussion as maintainers before we agree to merge this ⚠️ 

## Problem
We've had some issues regarding the order that the advert state is updated. In the current implementation we make certain assumptions about the ordering in which the `timeline` and `discontinuity` events are emitted. Wouldn't it be nice if we could just ignore the order and still emit the events in the same way?

## Solution
Implement an `update` method for the underlying `AdPlaybackState` that can be triggered from `timeline` and `discontinuity` regardless of the `reason` those methods give. This approach is significantly less complicated and doesn't require any knowledge of the hacks we've had to implement as workarounds for the different state problems we have found. 

### Test(s) added 
Not yet. I have a separate branch which begins to introduce an [AdvertState](https://github.com/novoda/no-player/blob/advert_loader_state_rework_testing/core/src/main/java/com/novoda/noplayer/internal/exoplayer/AdvertState.java) and [AdvertState tests](https://github.com/novoda/no-player/blob/advert_loader_state_rework_testing/core/src/test/java/com/novoda/noplayer/internal/exoplayer/AdvertStateTest.java) (name will probably change), where we pull all of the logic out of the `NoPlayerAdsLoader` and test it using just `position` info and nothing else 💥 

### But Ryan, did you test the real thing?
I spent a good while testing the demo we have for adverts and in a client application and couldn't find any issues with this new approach. The only scenario that I was unable to test was advert prepare errors or timeouts. 

### Paired with 
Nobody.


core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -179,39 +182,54 @@ public void onTimelineChanged(Timeline timeline, @Nullable Object manifest, int
...
-            handleAdvertStart();
-        }
+        updateState();
     }
 
-    private void handleAdvertStart() {
-        if (advertHasNotStarted()) {
-            adGroupIndex = player.getCurrentAdGroupIndex();
-            adIndexInGroup = player.getCurrentAdIndexInAdGroup();
+    private void updateState() {
```
Mecharyry: In this solution the `AdvertIndexInAdvertGroup` and the `isPlaying` determines how everything runs. In the event that `multiple calls` are made with the same index the new value will not be different and therefore will not retrigger events. If the index is different then we determine whether we were playing an advert in the last call and potentially finish the advert / advert break.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -39,8 +41,9 @@
 
     private NoPlayer.AdvertListener advertListener = NoOpAdvertListener.INSTANCE;
     private List<AdvertBreak> advertBreaks = Collections.emptyList();
-    private int adIndexInGroup = -1;
-    private int adGroupIndex = -1;
+    private boolean playingAdvert;
```
JozefCeluch: `isPlayingAdvert`?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -179,39 +182,54 @@ public void onTimelineChanged(Timeline timeline, @Nullable Object manifest, int
...
-            } else {
-                resetAdvertPosition();
-            }
+        if (advertFinished) {
+            adPlaybackState = adPlaybackState.withPlayedAd(previousAdvertGroupIndex, previousAdvertIndexInAdvertGroup);
+            updateAdPlaybackState();
+            notifyAdvertEnd(previousAdvertGroupIndex, previousAdvertIndexInAdvertGroup);
+        }
+
+        if (playingAdvert && advertIndexInAdvertGroup != previousAdvertIndexInAdvertGroup) {
```
JozefCeluch: Could we extract this condition to be `boolean advertStarted`? That way it would be symmetrical `advertFinished`

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -179,39 +182,54 @@ public void onTimelineChanged(Timeline timeline, @Nullable Object manifest, int
...
+            notifyAdvertStart(advertBreaks.get(advertGroupIndex));
```
JozefCeluch: Same thing here about making it symmetrical to `notifyAdvertEnd`. We could pass in `advertGroupIndex` and `advertIndexInAdvertGroup` (even though they are fields)

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: f7d649f

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: 72cc70e

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
Mecharyry: cdc57a7

Mecharyry:
Adding ❌ do not merge because we will test this in a client application with some proper QA before merging to develop.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -179,39 +182,57 @@ public void onTimelineChanged(Timeline timeline, @Nullable Object manifest, int
...
+        if (advertStarted) {
+            notifyAdvertStart(advertGroupIndex, advertIndexInAdvertGroup);
         }
     }
 
-    private boolean advertHasNotStarted() {
-        return player.isPlayingAd() && (adGroupIndex == -1 || adIndexInGroup == -1);
+    private void notifyAdvertEnd(int playedAdvertGroupIndex, int playedAdvertIndexInAdvertGroup) {
+        AdvertBreak advertBreak = advertBreaks.get(playedAdvertGroupIndex);
+        List<Advert> adverts = advertBreak.adverts();
```
zegnus: you could add a new method in the `advertBreak.advert(index)` so that you don't have to return the whole list

Mecharyry:
It replaces reason because the `onPositionDiscontinuity` doesn't deal with it anymore, it just ignores it and forwards to the update method.

Mecharyry:
We've had this tested in a client application and everything seems good to go!

PR #261, opened 2019-06-20
Key request executor

Mecharyry:
## Problem
We have a `LocalDrmSessionManager` which is super old and is used for starting a local Drm session. We can use the `DefaultDrmSessionManager` for this and move towards a `Drm` system where if the `KeySetId` is missing we can request another. 

## Solution
Rename `StreamingModularDrm` to `KeyRequestExecutor`, the idea being that in future the client can supply `KeySetId` AND this `KeyRequestExecutor` so that in the event that the `KeySetId` is invalid it will fetch another, which it cannot do at the moment. Force all `DrmSessionCreators` to talk in terms of the `DefaultDrmSession` because it can handle both `Streaming` and `Download` `DRM` just fine. 

This is all going on a feature branch at the moment because I want to test this solution in it's entirety before pushing to `develop`.

## Next steps
To create a `DrmSessionManager` based on whether the `KeySetId` has expired or not. In the case it is expired it will switch to the `Streaming` version. 

### Test(s) added 
No, just moving code around. 

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ProvisioningModularDrmCallback.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ProvisioningModularDrmCallback.java
@@ -4,18 +4,18 @@ import com.google.android.exoplayer2.drm.ExoMediaDrm;
 import com.google.android.exoplayer2.drm.MediaDrmCallback;
 import com.novoda.noplayer.drm.ModularDrmKeyRequest;
 import com.novoda.noplayer.drm.ModularDrmProvisionRequest;
-import com.novoda.noplayer.drm.StreamingModularDrm;
+import com.novoda.noplayer.drm.KeyRequestExecutor;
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ProvisioningModularDrmCallback.java
```diff
@@ -26,6 +26,6 @@
 
     @Override
     public byte[] executeKeyRequest(UUID uuid, ExoMediaDrm.KeyRequest request) throws Exception {
-        return streamingModularDrm.executeKeyRequest(new ModularDrmKeyRequest(request.getLicenseServerUrl(), request.getData()));
+        return keyRequestExecutor.executeKeyRequest(new ModularDrmKeyRequest(request.getLicenseServerUrl(), request.getData()));
```
zegnus: 👍

PR #262, opened 2019-06-20
DrmSession rework

Mecharyry:
## Problem
#261 We created a generic sort of `KeyRequestExecutor` and started using the `DefaultDrmSessionManager` in the hope that we could combine the two into the ULTIMATE `DrmSessionManagerCreator` 😂 

## Solution
Create a `DrmSessionManager` that will create a `DefaultDrmSessionManager` for which the `KeyRequestExecutor` can be specified by clients. For downloads this could just return the bytes of the `KeySetId` for example. 

This new `DrmSessionManager` will check for the `KeySetId`, if it is present it will check the expiry time, if it has more than 1 hour remaining of usage then it will use the `KeySetId` to play content. If the key is less than an hour, it will do nothing, which means it will perform a `KeyRequest` for the session to perform playback. 

I've removed the `DrmHandler` because this is now replaced by the `KeyRequestExecutor`, if clients want to use Drm they need to supply a means for retrieving keys in order for content to be decrypted.

### Paired with 
Nobody.


core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
```diff
@@ -0,0 +1,90 @@
...
+import com.novoda.noplayer.drm.KeyRequestExecutor;
+import com.novoda.noplayer.internal.drm.provision.ProvisionExecutor;
+import com.novoda.noplayer.model.KeySetId;
+
+import java.util.HashMap;
+import java.util.concurrent.TimeUnit;
+
+import androidx.annotation.Nullable;
+
+class ExoPlayerDrmSessionCreator implements DrmSessionCreator {
```
Mecharyry: This is the important class which created a unified `DrmSessionManager` for downloads and streaming based on the `KeySetId`.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
```diff
@@ -0,0 +1,90 @@
...
+
+        FrameworkMediaDrm frameworkMediaDrm = mediaDrmCreator.create(WIDEVINE_MODULAR_UUID);
+        DefaultDrmSessionManager<FrameworkMediaCrypto> drmSessionManager = new DefaultDrmSessionManager<>(
+                WIDEVINE_MODULAR_UUID,
+                frameworkMediaDrm,
+                mediaDrmCallback,
+                NO_OPTIONAL_PARAMETERS
+        );
+
+        if (keySetId != null) {
```
Mecharyry: If we have a `keySetId` then we want to try and restore it.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
```diff
@@ -0,0 +1,90 @@
...
+            OfflineLicenseHelper<FrameworkMediaCrypto> offlineLicenseHelper = OfflineLicenseHelper.newWidevineInstance(
+                    "",
+                    null
+            );
+
+            byte[] rawKeySetId = keySetId.asBytes();
+            Pair<Long, Long> licenseDurationRemainingSec = offlineLicenseHelper.getLicenseDurationRemainingSec(rawKeySetId);
+            Long first = licenseDurationRemainingSec.first;
+            if (first > TimeUnit.HOURS.toSeconds(1)) {
+                drmSessionManager.setMode(DefaultDrmSessionManager.MODE_QUERY, rawKeySetId);
```
Mecharyry: In the event that we still have 1 hour remaining we will set this mode, so the `DrmManager` knows to use the `KeySetId`.

Mecharyry: In the event that it fails or it has expired, we do nothing, which makes the `DrmSessionManager` use the default mode, which will execute a key request.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
```diff
@@ -0,0 +1,90 @@
...
+        drmSessionManager.addListener(handler, eventListener);
+
+        return drmSessionManager;
+    }
+
+    private void checkExpiryOfKeySetId(DefaultDrmSessionEventListener eventListener,
+                                       DefaultDrmSessionManager<FrameworkMediaCrypto> drmSessionManager) {
+        try {
+            OfflineLicenseHelper<FrameworkMediaCrypto> offlineLicenseHelper = OfflineLicenseHelper.newWidevineInstance(
+                    "",
```
zegnus: can you explain those params? potentially extracting them into meaningful names

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
```diff
@@ -0,0 +1,90 @@
...
+                    null
+            );
+
+            byte[] rawKeySetId = keySetId.asBytes();
+            Pair<Long, Long> licenseDurationRemainingSec = offlineLicenseHelper.getLicenseDurationRemainingSec(rawKeySetId);
+            Long first = licenseDurationRemainingSec.first;
```
zegnus: is `Long first` the remaining time? if so, rename it :)

core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
```diff
@@ -61,31 +65,39 @@ public PlayerBuilder withWidevineClassicDrm() {
...
      *
-     * @param drmType    {@link DrmType}
-     * @param drmHandler {@link DrmHandler}
+     * @param drmType            {@link DrmType}
+     * @param keyRequestExecutor {@link KeyRequestExecutor}
      * @return {@link PlayerBuilder}
      * @see NoPlayer
      */
-    public PlayerBuilder withDrm(DrmType drmType, DrmHandler drmHandler) {
+    public PlayerBuilder withDrm(DrmType drmType, KeyRequestExecutor keyRequestExecutor, KeySetId keySetId) {
```
juankysoriano: So, when creating we pass a current `KeySetId` and a `KeyRequestExecutor` to  be executed when expired or about to explire, being `no-player` the one determining when `about to expire` evaluates to `true`.

Is it ok that we do it this way? wouldn't different clients of the library have different requirements on regards to what `about to expire` is?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
```diff
@@ -0,0 +1,90 @@
...
+            if (first > TimeUnit.HOURS.toSeconds(1)) {
+                drmSessionManager.setMode(DefaultDrmSessionManager.MODE_QUERY, rawKeySetId);
+            }
+
+        } catch (UnsupportedDrmException | DrmSession.DrmSessionException e) {
```
zegnus: maybe for a next task, let's check if we actually handle properly those types of exceptions

core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
Mecharyry: This is actually one of the discussions that I had with @zegnus. I wasn't 100% sure if this was the best approach because we could move the `KeySetId` logic higher.

We could have it so they logic in this class is split for `Downloads` and `Streaming`, almost how it was before. Having the client decide when to use one or the other. It means the client would need to import some helper that is able to determine the expiry time of the `KeySetId` and choose appropriately.

I personally don't mind either approach, this was the one we settled on for now. cc @zegnus

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
```diff
@@ -0,0 +1,90 @@
...
+    private void checkExpiryOfKeySetId(DefaultDrmSessionEventListener eventListener,
+                                       DefaultDrmSessionManager<FrameworkMediaCrypto> drmSessionManager) {
+        try {
+            OfflineLicenseHelper<FrameworkMediaCrypto> offlineLicenseHelper = OfflineLicenseHelper.newWidevineInstance(
+                    "",
+                    null
+            );
+
+            byte[] rawKeySetId = keySetId.asBytes();
+            Pair<Long, Long> licenseDurationRemainingSec = offlineLicenseHelper.getLicenseDurationRemainingSec(rawKeySetId);
```
Mecharyry: I should point out that this is a blocking call @zegnus @juankysoriano but because it is querying an existing `KeySetId` it should be instantaneous, unless the `DrmEngine` locks up for any reason.

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
Mecharyry: As per the docs on the `onDrmSessionManagerError`: 
```
  /**
   * Called when a drm error occurs.
   *
   * <p>This method being called does not indicate that playback has failed, or that it will fail.
   * The player may be able to recover from the error and continue. Hence applications should
   * <em>not</em> implement this method to display a user visible error or initiate an application
   * level retry ({@link Player.EventListener#onPlayerError} is the appropriate place to implement
   * such behavior). This method is called to provide the application with an opportunity to log the
   * error if it wishes to do so.
   *
   * @param error The corresponding exception.
   */
```

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
```diff
@@ -0,0 +1,90 @@
...
+            Long remainingLicenseTimeInSeconds = licenseDurationRemainingSec.first;
+            Long playbackTimeInSeconds = licenseDurationRemainingSec.second;
+            if (remainingLicenseTimeInSeconds > playbackTimeInSeconds) {
```
Mecharyry: @juankysoriano this might alleviate your concerns. I remembered that we can grab the `playbackTime` from the `keySetId`. I'll need to check this to make 100% sure that this works as expected though. Not sure if we need a little bit offset too? @zegnus you might be interested in this too.

core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
Mecharyry: Also -> b3ddfbe

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
Mecharyry: b3ddfbe

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
Mecharyry: b8088ba

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
Mecharyry: b3ddfbe

core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
juankysoriano: I think it's client responsibility to handle the renewal of the keys, as different DRM systems could have different requirements.

For now we can leave it like this, we should triple think this

core/src/main/java/com/novoda/noplayer/internal/exoplayer/drm/ExoPlayerDrmSessionCreator.java
juankysoriano: This is definitively a bit better than a static `1 hour`. Although I am still a bit worried that "trying to solve too much" for the client could create some undesired side effects on some implementations. 

For now we can go with this but we should try to explore a bit further about what's the best way to deal with this on general terms

core/src/main/java/com/novoda/noplayer/PlayerBuilder.java
zegnus: While I agree on that potentially the decision should be made by the client, after talking with Ryan we decided to go for this approach. The reason is that this piece of code isolates the client from lots of complexity and potential bugs, while at the same time agreeing that the logic is not a fallback but a default behaviour. If the client passes keys, use it, if they are expired, delegate to a requestKeys method (set up by the client).

We can always go through this again in a call

PR #265, opened 2019-06-24
Codec capabilities

Mecharyry:
## Problem
We've been testing the #263 DRM Rework and we discovered that on some devices we were seeing codec failures. Which is super strange when we include `secure` and `non-secure` codecs AND also fallback when there are any errors. 🤔 

I had a little look into how the renderers select codecs and it seems that `ExoPlayer` is a little too optimistic 😄 It picks the first decoder and roles with it even if it states that it cannot support a given format. I posted a message in the `ExoPlayer` Slack and it seems this was intentional because `hardware` decoders normally understate what they can playback, which sometimes leads to falling back to a software decoder unnecessarily. They did acknowledge the problem and it seems they have a fix for this in the `dev` branch but it hasn't been released yet. https://github.com/google/ExoPlayer/blob/dev-v2/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java#L376-L404

## Solution
Copy the code from: 
https://github.com/google/ExoPlayer/blob/dev-v2/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java#L385 and https://github.com/google/ExoPlayer/blob/dev-v2/library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java#L342-L364 to sort the codecs until the version supporting the changes is released. 

## Testing
I've tested this on a client application using a problematic device in question and all seems good again.

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/InternalMediaCodecUtil.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/InternalMediaCodecUtil.java
@@ -1,0 +1,21 @@
+package com.novoda.noplayer.internal.exoplayer;
+
+import com.google.android.exoplayer2.mediacodec.MediaCodecInfo;
+import com.google.android.exoplayer2.mediacodec.MediaCodecUtil;
+
+import java.util.List;
+
+class InternalMediaCodecUtil {
+
+    List<MediaCodecInfo> getDecoderInfos(
+            String mimeType,
+            boolean requiresSecureDecoder,
+            boolean requiresTunnelingDecoder
+    ) throws MediaCodecUtil.DecoderQueryException {
+        return MediaCodecUtil.getDecoderInfos(mimeType, requiresSecureDecoder, requiresTunnelingDecoder);
+    }
+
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/InternalMediaCodecUtil.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * Stably sorts the provided {@code list} in-place, in order of decreasing score.
+     */
+    private static <T> void sortByScore(List<T> list, final ScoreProvider<T> scoreProvider) {
+        Collections.sort(list, new Comparator<T>() {
+            @Override
+            public int compare(T a, T b) {
+                return scoreProvider.getScore(b) - scoreProvider.getScore(a);
```
JozefCeluch: so if we get `-1` for `a` and `1` for `b` wouldn't that put `a` before `b`?

Mecharyry: I just copied the class and converted it to anonymous classes, I haven't touched it otherwise.

PR #267, opened 2019-07-03
Content position and duration

Mecharyry:
## Problem
We have `mediaDurationInMillis` and `playheadPositionInMillis` which can return very different values based on the underlying state of the player 😬 Ideally we should guarantee that methods will have predictable behaviour that does not require extensive documentation.  

## Solution
Introduce a `contentDurationInMillis` and `contentPositionInMillis` that **only** deal with the content. Adverts should be considered separately. This should help to avoid lots of checks for `adverts` and `content` when we already know it is safe to call these methods. 

### Test(s) added 
Yes where necessary.

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/PlayerState.java
+++ b/core/src/main/java/com/novoda/noplayer/PlayerState.java
@@ -86,6 +86,14 @@ public interface PlayerState {
     long mediaDurationInMillis();
 
     /**
+     * Duration of the content that is being played.
+     * Use {@link PlayerState#mediaDurationInMillis} for duration of an advert when one is playing.
+     *
+     * @return duration of the content that is being played.
+     */
+    long contentDurationInMillis();
+
+    /**
      * Buffered percentage of the video that is being played.
      * <p>
      * If the player is playing an advert break then this value will represent a buffer of each individual advert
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/novoda/noplayer/PlayerState.java
```diff
@@ -75,6 +85,14 @@
      */
     long mediaDurationInMillis();
```
danybony: So this is going to be the same as `contentDurationInMillis` when playing main content and changing when playing an ad, right? (same for `playheadPositionInMillis`?)

Mecharyry: Yea, which makes it annoying for clients to use because we can't guarantee the behaviour. Whereas the new methods will only talk about content.

core/src/main/java/com/novoda/noplayer/PlayerState.java
```diff
@@ -65,6 +65,16 @@
      */
     long playheadPositionInMillis();
```
JozefCeluch: Should we deprecate this old method then? It seems like having a _convenience_ method like this will only lead to subtle errors.

Mecharyry: If it is ok with you I'd rather deprecate it when we have all of the advert specific ones in place as well. As it is we have a mismatch 😄

JozefCeluch: That's fine, I thought having advert break methods is enough, but it's true that we should probably expose advert-specific methods too (... at some point 😄)

PR #268, opened 2019-07-03
Allow require secure decoders

zegnus:
## Problem

A client needs to set up in the player builder if it requires secure decoders

## Solution

Added a requires secure decoder option in the builder.
Also in the Track object I have added additional information:
- Codec name
- Renderer capabilities for the current track

### Test(s) added 

Yes

### Screenshots

No UI changes

### Paired with 

Nobody


core/src/main/java/com/novoda/noplayer/internal/exoplayer/MediaCodecVideoRendererWithSimplifiedDrmRequirement.java
```diff
@@ -54,13 +57,18 @@
...
     @Override
     protected List<MediaCodecInfo> getDecoderInfos(MediaCodecSelector mediaCodecSelector,
                                                    Format format,
                                                    boolean requiresSecureDecoder) throws MediaCodecUtil.DecoderQueryException {
-        return getDecoderInfos(mediaCodecSelector, format, format.drmInitData != null, getCodecNeedsEosPropagation());
+        return getDecoderInfos(mediaCodecSelector, format, requiresSecureDecoder(format), getCodecNeedsEosPropagation());
+    }
+
+    private boolean requiresSecureDecoder(Format format) {
+        return format.drmInitData != null && requiresSecureDecoder;
```
zegnus: this keeps the previous behaviour in place

core/src/main/java/com/novoda/noplayer/internal/exoplayer/MediaCodecVideoRendererWithSimplifiedDrmRequirement.java
```diff
@@ -70,27 +78,38 @@
             boolean requiresSecureDecoder,
             boolean requiresTunnelingDecoder)
             throws MediaCodecUtil.DecoderQueryException {
-        List<MediaCodecInfo> decoderInfos =
-                mediaCodecSelector.getDecoderInfos(
-                        format.sampleMimeType, requiresSecureDecoder, requiresTunnelingDecoder);
+        List<MediaCodecInfo> decoderInfos = mediaCodecSelector.getDecoderInfos(
```
zegnus: no change in this class, just some refactoring so that we can understand the code better

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/ExoPlayerTrackSelector.java
```diff
@@ -25,8 +26,26 @@ private ExoPlayerTrackSelector(DefaultTrackSelector trackSelector, RendererTrack
     }
 
     TrackGroupArray trackGroups(TrackType trackType, RendererTypeRequester rendererTypeRequester) {
-        Optional<Integer> audioRendererIndex = rendererTrackIndexExtractor.extract(trackType, mappedTrackInfoLength(), rendererTypeRequester);
-        return audioRendererIndex.isAbsent() ? TrackGroupArray.EMPTY : trackInfo().getTrackGroups(audioRendererIndex.get());
+        Optional<Integer> rendererIndex = rendererTrackIndexExtractor.extract(trackType, mappedTrackInfoLength(), rendererTypeRequester);
+        return rendererIndex.isAbsent() ? TrackGroupArray.EMPTY : trackInfo().getTrackGroups(rendererIndex.get());
+    }
+
+    RendererCapability rendererCapability(TrackType trackType, int groupIndex, int trackIndex, RendererTypeRequester rendererTypeRequester) {
```
zegnus: new renderer capabilities information

core/src/main/java/com/novoda/noplayer/internal/exoplayer/mediasource/ExoPlayerVideoTrackSelector.java
```diff
@@ -41,18 +41,20 @@ public boolean selectVideoTrack(PlayerVideoTrack videoTrack, RendererTypeRequest
...
-                        formatIndex,
+                        trackIndex,
                         format.id,
                         contentType,
                         format.width,
                         format.height,
                         (int) format.frameRate,
-                        format.bitrate
+                        format.bitrate,
+                        format.codecs,
```
zegnus: this will add the track content codec name so that the client can know about it

core/src/main/java/com/novoda/noplayer/model/RendererCapability.java
```diff
@@ -0,0 +1,28 @@
+package com.novoda.noplayer.model;
+
+import com.google.android.exoplayer2.RendererCapabilities;
```
Mecharyry: This can't be exposed here. All exoplayer references need to be moved to the internal package. We shouldn't be exposing this. If we have a `RendererCapability` there should be a class that knows how to transform to it from whatever the underlying player is.

core/src/main/java/com/novoda/noplayer/model/RendererCapability.java
```diff
@@ -0,0 +1,28 @@
...
+
+public enum RendererCapability {
```
Mecharyry: I'd prefer if this was called `Support` tbh, it's the level of `support` that a given track has.

demo/src/main/java/com/novoda/demo/MainActivity.java
```diff
@@ -132,7 +132,7 @@ public void onLicenseDownloaded(byte[] license) {
                 player = playerBuilder.build(getApplicationContext());
 
                 demoPresenter = new DemoPresenter(controllerView, player, player.getListeners(), playerView);
-                dialogCreator = new DialogCreator(getApplicationContext(), player);
+                dialogCreator = new DialogCreator(player);
```
Mecharyry: Why are we not passing in the context through the constructor anymore?

zegnus: ah yes, because it blows up. Current develop is broken and when you click on the video selector the application crashes. A Dialog needs an Activity.

Mecharyry: > A dialog needs an activity

Dialogs can be attached using a context, otherwise the builder would have complained. Probably it broke when we moved things around.

zegnus: well yeah but I mean the Activity context

core/src/main/java/com/novoda/noplayer/model/RendererCapability.java
zegnus: it has this name internally, I've changed

core/src/main/java/com/novoda/noplayer/model/PlayerVideoTrack.java
```diff
@@ -12,9 +12,20 @@
     private final int height;
     private final int fps;
     private final int bitrate;
-
-    @SuppressWarnings("checkstyle:ParameterNumber") // TODO group parameters into classes
-    public PlayerVideoTrack(int groupIndex, int formatIndex, String id, ContentType contentType, int width, int height, int fps, int bitrate) {
+    private final String codecName;
+    private final RendererSupport rendererCapability;
```
Mecharyry: This is still referencing a renderer which is an internal concept of exoplayer. `Support` implies it is only about this model and this model alone, without leaking exoplayer into our domain. Also, this variable name is still `rendererCapability`.

PR #269, opened 2019-07-04
Add track decoder association

zegnus:
## Problem

We do not know which decoder is going to be used by the engine per track.

## Solution

Surface that information using a singleton. I have tried to do it without a singleton but I didn't find a nice and performance way to do it.

I have also provided a list of decoders that *actually* supports the required format, instead of providing decoders that the engine already says that it is not supported.

### Test(s) added 

All tested

### Screenshots

Screenshot_1562256914

### Paired with 

Nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/InternalMediaCodecUtil.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/InternalMediaCodecUtil.java
@@ -51,7 +51,7 @@ class InternalMediaCodecUtil {
 
     /**
      *
-     * Returns a list of decoders that supports the procided format
+     * Returns a list of decoders that supports the provided format
      */
     static List<MediaCodecInfo> getOnlySupportedDecoderInfos(List<MediaCodecInfo> decoderInfos, Format format) {
         List<MediaCodecInfo> onlySupported = new ArrayList<>();
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/InternalMediaCodecUtil.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/InternalMediaCodecUtil.java
@@ -50,6 +50,25 @@ class InternalMediaCodecUtil {
     }
 
     /**
+     *
+     * Returns a list of decoders that supports the procided format
+     */
+    static List<MediaCodecInfo> getOnlySupportedDecoderInfos(List<MediaCodecInfo> decoderInfos, Format format) {
+        List<MediaCodecInfo> onlySupported = new ArrayList<>();
+        for (MediaCodecInfo decoderInfo : decoderInfos) {
+            try {
+                if (decoderInfo.isFormatSupported(format)) {
+                    onlySupported.add(decoderInfo);
+                }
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/MediaCodecVideoRendererWithSimplifiedDrmRequirement.java
```diff
@@ -110,6 +113,26 @@ private boolean requiresSecureDecoder(Format format) {
...
+    private static void saveTrackCodecMapping(String trackCodecName, List<MediaCodecInfo> decoderInfos) {
+        PlayerVideoTrackCodecMapping.getInstance().addTrackCodec(trackCodecName, codecName(decoderInfos));
+    }
+
+    private static String codecName(List<MediaCodecInfo> decoderInfos) {
+        return decoderInfos.isEmpty() ? "no codec available" : decoderInfos.get(0).name;
+
+    }
+
+    private static void printDecoderInfos(String codecs, List<MediaCodecInfo> decoderInfos) {
```
zegnus: this is useful for debug purposes

core/src/main/java/com/novoda/noplayer/internal/exoplayer/InternalMediaCodecUtil.java
```diff
@@ -49,6 +49,25 @@ public int getScore(MediaCodecInfo decoderInfo) {
         return decoderInfos;
     }
 
+    /**
+     *
+     * Returns a list of decoders that supports the procided format
```
Mecharyry:
```suggestion
     * Returns a list of decoders that supports the provided format
```

core/src/main/java/com/novoda/noplayer/model/PlayerVideoTrackCodecMapping.java
```diff
@@ -0,0 +1,30 @@
...
+ * This class holds an association betwen the codec being used by a particular track
+ * and the decoder that the engine assigns to this codec.
+ *
+ * Bear in mind that this association is provisional and there is no guarantee that
+ * the engine will not fallback to another decoder. This is due to that usually several
+ * decoders are provided for a particual track. If the first decoder fails to initialise
+ * then the engine will choose another decoder. This information is not easely accessible
+ * and so we only provide the first association that the engine does.
+ */
+public final class PlayerVideoTrackCodecMapping {
```
Mecharyry: When is this cleared?

zegnus: I was thinking about that. Currently is not, it will be overridden by the next play. So that when new tracks comes in and new decoders are associated, the map is updated. It could be that the new stream has fewer tracks and then the map has invalid extra tracks, but if the request from the map comes from the actual tracks (that it is from the No-Player API point of view) then we should be fine. The information is attached to the VideoTrack.

Mecharyry: Ah I see. Yea this is not an issue then. 👍

PR #271, opened 2019-07-19
Secure and unsupported codecs

Mecharyry:
## Problem
We want to be able to remove unsupported codecs from the list of available codecs when preparing a track for playback. Additionally we want to force all tracks that are considered to be HD to use `secure` codecs. 

## Solution
Introduce two new parameters to the `Options` interface:

`withSecureDecoderRequiredForHD(int hdQualityBitrateThreshold)` - above which the track will remove all codecs that are not secure. 

`withUnsupportedVideoDecoders(List<String> unsupportedVideoDecoders)` - the names of the codecs that should be removed from all tracks.

### Test(s) added 
Yes. All of the changes related to stripping out codecs is done on the `InternalMediaCodecUtil` so these new methods have been tested.

### Screenshots
Shows removing the codec from the HD track.

device-2019-07-19-152254

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/Options.java
+++ b/core/src/main/java/com/novoda/noplayer/Options.java
@@ -16,6 +16,7 @@ public class Options {
     private final Optional<Long> initialPositionInMillis;
     private final Optional<Long> initialAdvertBreakPositionInMillis;
     private final List<String> unsupportedVideoDecoders;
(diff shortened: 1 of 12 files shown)


Mecharyry:
For those wondering that chat head is my mother 😂

PR #275, opened 2019-08-05
NO-272/Advert Tests

Mecharyry:
## Problem
When creating the `NoPlayerAdsLoader` we were a little concerned that we had very few tests for it 😬 and that it was generally difficult to test. 

## Solution
Extract logic related adverts to a more testable component `AdvertState` and add tests 😄 

### Paired with 
Nobody.

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
@@ -25,7 +25,7 @@ import androidx.annotation.Nullable;
 
 // Not much we can do, orchestrating adverts is a lot of work.
 @SuppressWarnings("PMD.GodClass")
-public class NoPlayerAdsLoader implements AdsLoader, Player.EventListener, AdvertView.AdvertInteractionListener {
+public class NoPlayerAdsLoader implements AdsLoader, Player.EventListener, AdvertView.AdvertInteractionListener, AdvertState.Callback {
 
     private final AdvertsLoader loader;
     private final Handler handler;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/NoPlayerAdsLoader.java
```diff
@@ -147,8 +143,8 @@ public void stop() {
             loadingAds.cancel();
             loadingAds = null;
         }
-        if (adPlaybackState != null && player != null && player.isPlayingAd()) {
-            adPlaybackState = adPlaybackState.withAdResumePositionUs(TimeUnit.MILLISECONDS.toMicros(player.getCurrentPosition()));
+        if (adPlaybackState != null && player != null && advertState != null) {
```
Mecharyry: The guards are essentially based on the output. Here, because we are stopping, we need to store the position (player), notify the player (AdPlaybackState) and reset the advert state.

core/src/test/java/com/novoda/noplayer/internal/exoplayer/AdvertStateTest.java
```diff
@@ -0,0 +1,307 @@
...
+        advertState.update(IS_NOT_PLAYING_ADVERT, UNSET_ADVERT_BREAK_INDEX, UNSET_ADVERT_INDEX);
+
+        then(callback).shouldHaveZeroInteractions();
+    }
+
+    @Test
+    public void notifiesStartOfAdvertBreak_andAdvert_whenTransitioningToAdverts() {
+        advertState.update(IS_PLAYING_ADVERT, 0, 0);
+
+        then(callback).should().onAdvertBreakStart(FIRST_ADVERT_BREAK);
```
JozefCeluch: would it make sense to use `Mockito.inOrder()` for these assertions to ensure that the order doesn't get changed by accident?

Mecharyry: Huh weird, yea all of these tests should have `inOrder` to them 😬 because the order does actually matter. Thanks for noticing @JozefCeluch I don't know how I managed to not include it 😭

PR #277, opened 2019-09-23
Support user defined captions

pablisco:
## Problem

We need to support user defined styles for closed captions on the player as the current implementation disregards the user settings and uses the ones provided by the manifest of the video.

## Solution

- Create a component (`SystemCaptionPreferences`), used by `SubtitlePainter` to determine how to paint the subtitles.
- This component creates a `SubtitleStyle` with the relevant style that the user has defined based on `CaptionStyleCompat`.
- When there is no user definition (because the settings are disabled), the device is below API 19 (when the accessibility API was introduced) or the user of the library has explicitly disable it (see README for more details) then `SystemCaptionPreferences` produces an instance of `NoSubtitlesStyle` which returns the fallback values.

In `SubtitlePainer` we were before doing this:

```Java
this.backgroundColor = Color.BLACK;
```
but now instead we do the following:
```Java
this.backgroundColor = subtitlesStyle.backgroundColorOr(Color.BLACK);
```

So, if captions are disabled, then the default `BLACK` colour will be set.

The properties that we are using are:

- Background colour: The colour for the box where the subtitles are drawn
- Foreground colour: The colour of the text for the given subtitles.
- Window colour: This is a feature only available from API 21 and it's a colour set around the box where the text is drawn
- Text Scale: Used to make the text bigger
- Typeface: Used for the style of the text
- Edge type and colour: Used to add decorations to the text like edges or shadows

All colours take into account alpha.

#### Note:

The style is gathered each time we need it as it could change at any time since it's a setting that comes from the system and we have no control over it or a way to observe changes.

#### Note 2:

We've used reflexion for testing to avoid unnecessary collaborators in the code. Tried to do it using abstractions but it would have ended with 3 times the number of classes and very small tests that didn't test anything 😅 

#### Bonus: Testing

We had an issue regarding testing this `mpd` manifest as it doesn't require 

https://storage.googleapis.com/content-samples/multi-audio/manifest.mpd

So we updated the launching screen to be able to disable content protection so this url can be used to testing.

PS: also fixed a problem with hiding track selector when there is only one available.

### Test(s) added 

Added test for `SystemCaptionPreferences`

### Screenshots

Using these settings:

image

It would behave like this:

| Before | After |
| ------ | ----- |
| image | image |

### Paired with 

@Mecharyry


core/src/test/java/com/novoda/noplayer/subtitles/SystemCaptionPreferencesTest.java
```diff
@@ -0,0 +1,289 @@
...
+            when(style.hasEdgeType()).thenReturn(true);
+            setFinalField(style, "edgeColor", SYSTEM_EDGE_COLOR);
+            when(style.getTypeface()).thenReturn(SYSTEM_TYPEFACE);
+            return style;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+}
```
juankysoriano: `No new line at the end of file`

core/src/test/java/com/novoda/noplayer/subtitles/SystemCaptionPreferencesTest.java
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        TestCase captionsLocallyDisabled() {
+            this.captionsLocallyEnabled = false;
+            return this;
+        }
+
+        @NonNull
+        @Override
+        public String toString() {
```
juankysoriano: :+1:

demo/src/main/java/com/novoda/demo/LandingActivity.java
```diff
@@ -23,6 +23,7 @@ protected void onCreate(Bundle savedInstanceState) {
         final EditText mpdAddress = findViewById(R.id.mpd_address);
         final EditText licenseAddress = findViewById(R.id.license_address);
         final CheckBox downloadLicense = findViewById(R.id.download_license);
+        final CheckBox useContentProtection = findViewById(R.id.use_content_proteciton);
```
juankysoriano: What's this stuff about `contentProtection` and how does it relates to user defined caption style?

core/src/test/java/com/novoda/noplayer/subtitles/SystemCaptionPreferencesTest.java
```diff
@@ -0,0 +1,289 @@
...
+            PRE_KITKAT_TEST_CASE,
+            KITKAT_TEST_CASE,
+            POST_LOLLIPOP_TEST_CASE,
+            DISABLED_TEST_CASE,
+            LOCALLY_DISABLED_TEST_CASE
+        };
+    }
+
+    @Rule
+    public final StaticMutationRule mutations = new StaticMutationRule();
```
juankysoriano: It's good that we have made this a `rule` but I am under the impression that we are missusing it. We only take the advantage of the `after()` routine, but leaving the `before` to be implemented by the user of the rule. 

Can we pass the different mutations we want to perform, as a list of `functions` on the constructor and let them be applied on `before()`?

README.md
```diff
@@ -93,6 +93,13 @@ repositories {
 
 You can find the latest snapshot version following this [link](https://bintray.com/novoda-oss/snapshots/no-player/_latestVersion).
 
+
+## Accesibility Captions
+
+Accessibility settings set by the user are taken into account when displaying Close Captions **by default**.
```
Mecharyry:
```suggestion
Accessibility settings set by the user are taken into account when displaying Closed Captions **by default**.
```

It's Closed Captions 😄

core/src/main/java/com/novoda/noplayer/PlayerView.java
```diff
@@ -19,4 +19,12 @@
 
     void setSubtitleCue(TextCues textCues);
 
+    /**
+     * whether to use the styles set by the user in the accesibility settings instead of the style
```
Mecharyry:
```suggestion
     * Whether to use the styles set by the user in the accessibility settings instead of the style
```
Couple of typos

core/src/main/java/com/novoda/noplayer/PlayerView.java
```diff
@@ -19,4 +19,12 @@
...
+     * that comes from the video's subtitle cues.
+     *
+     * Default: Enabled (true)
+     */
+    void setAccessibilityCaptionsStyleEnabled(boolean enabled);
```
Mecharyry: This doesn't look like a true javadoc. Normally it labels the params as well.

core/src/main/java/com/novoda/noplayer/subtitles/KitkatSubtitlesStyle.java
```diff
@@ -0,0 +1,59 @@
...
+
+import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
+
+import com.google.android.exoplayer2.text.CaptionStyleCompat;
+
+import static android.os.Build.VERSION_CODES.KITKAT;
+
+@RequiresApi(api = KITKAT)
+class KitkatSubtitlesStyle implements SubtitlesStyle {
```
Mecharyry: Should be `KitKat` 😂 I'm being pedantic because KitKat is two words 👅

core/src/main/java/com/novoda/noplayer/subtitles/NoSubtitlesStyle.java
```diff
@@ -0,0 +1,45 @@
+package com.novoda.noplayer.subtitles;
+
+import android.graphics.Typeface;
+
+import androidx.annotation.Nullable;
+
+class NoSubtitlesStyle implements SubtitlesStyle {
```
Mecharyry: Is it a `NoSubtitlesStyle`? `FallbackOnlySubtitlesStyle`?

core/src/test/java/com/novoda/noplayer/test/utils/StaticMutationRule.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.rules.ExternalResource;
+
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.novoda.noplayer.test.utils.ReflectionFinalMutationUtils.*;
+
+/**
+ * Allows mutating constants which are restored after teh test is finished.
```
Mecharyry:
```suggestion
 * Allows mutating constants which are restored after the test is finished.
```

core/src/main/java/com/novoda/noplayer/SubtitleView.java
```diff
@@ -23,6 +23,7 @@
     private final List<SubtitlePainter> painters;
 
     private TextCues textCues;
+    private boolean accessibilityCaptionsStyleEnabled = true;
```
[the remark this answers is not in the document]
pablisco: I thought about this and it makes sense to be nice to users that require accessibility settings by default. There is an option to disable it but should encourage positive behaviour towards accessibility.

core/src/main/java/com/novoda/noplayer/subtitles/SystemCaptionPreferences.java
```diff
@@ -0,0 +1,48 @@
...
+
+import static android.os.Build.VERSION_CODES.KITKAT;
+import static com.google.android.exoplayer2.text.CaptionStyleCompat.createFromCaptionStyle;
+import static java.util.Objects.requireNonNull;
+
+public class SystemCaptionPreferences {
+
+    private final Context context;
+
+    private boolean enabled = true;
```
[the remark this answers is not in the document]
pablisco: Same as before

core/src/test/java/com/novoda/noplayer/subtitles/SystemCaptionPreferencesTest.java
pablisco: I thought about doing that but we are using Java 1.6 so we can't use lambdas, so this would make it harder/uglier to use :( Maybe if we add Kotlin in the future 😅
The important thing is to clear out the static state so I think this fulfils it's purpose.

One thing I thought of (but didn't want to complicate things) was to have a type that we can pass into the constructor. Happy to have a look at doing something like that:

```Java
new StaticMutationRule(
  mutate(Build.VERSION.class, "SDK_INT", testCase.sdkLevel),
  mutate(Util.class, "SDK_INT", testCase.sdkLevel)
);
```
Where `mutate` is a SFM that creates a class that has an private constructor

core/src/main/java/com/novoda/noplayer/subtitles/KitkatSubtitlesStyle.java
pablisco: Doing this change, however it depends if we spell it like `Kit Kat` or `KitKat`. If it's the second then it would follow the similar rules for acronyms, `HTTP -> Http`.

And the official website spells it as one word 🙃 

image

I like being pedantic too 😅

core/src/main/java/com/novoda/noplayer/subtitles/NoSubtitlesStyle.java
pablisco: Yeah, I wasn't sure as the name is a bit ambiguous as it could refer to the style when there is no subtitles. They are not all fallback as text size just multiplies by 1. 
I'll rename it to `NoOpSubtitlesStyle`

README.md
```diff
@@ -93,6 +93,13 @@ repositories {
...
+
+It's possible to disable this behaviour by calling `setAccessibilityCaptionsStyleEnabled(false)` on `PlayerView`.
```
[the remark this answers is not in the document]
pablisco: https://github.com/novoda/no-player/pull/277/commits/e69d0b1130499351332c09915e8e2e1fd4b1058a

core/src/test/java/com/novoda/noplayer/subtitles/SystemCaptionPreferencesTest.java
pablisco: https://github.com/novoda/no-player/pull/277/commits/01eb575af6074460e4e7cdfdad1f73dcc0b78ea5

core/src/test/java/com/novoda/noplayer/subtitles/SystemCaptionPreferencesTest.java
pablisco: https://github.com/novoda/no-player/pull/277/commits/3b1719a5b588461d6e26ce212ab39658f5689330

core/src/main/java/com/novoda/noplayer/PlayerView.java
pablisco: https://github.com/novoda/no-player/pull/277/commits/746cee0bf23c52c84dce2d44f832fa643a89d6ff

core/src/main/java/com/novoda/noplayer/subtitles/KitkatSubtitlesStyle.java
pablisco: https://github.com/novoda/no-player/pull/277/commits/e94a6beadc23aad21c86cbdfbd4f1c3fb93708d4

core/src/main/java/com/novoda/noplayer/subtitles/NoSubtitlesStyle.java
pablisco: https://github.com/novoda/no-player/pull/277/commits/5c139883cb015dcbf3a89eb24283d48e5fa52713

PR #280, opened 2019-10-07
Add max vide size constrain

zegnus:
## Problem

We need to add the capability of limiting the maximum video size that the ExoPlayer will allow for playback

## Solution

Expose a public method for setting the max video size that will connect to the native track selection max video size in the ExoPlayer. Also for this purpose, I have created a `Dimension` class in order to encapsulate width and hight.

I have also updated the project dependencies.

`toString`, `hashCode` and `equals` have been generated using the `IDE` code generation feature

### Test(s) added 

Tested in the demo application. While doing this I have noticed that the checkboxes were not working as expected, actually, they were implemented in reverse (checked was no limit, unchecked with a limit). I have fixed the HD checkbox and added an additional one.

### Screenshots

Screenshot_1570442907

### Paired with 

Nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/NoPlayer.java
+++ b/core/src/main/java/com/novoda/noplayer/NoPlayer.java
@@ -2,10 +2,13 @@ package com.novoda.noplayer;
 
 import android.net.Uri;
 
+import androidx.annotation.FloatRange;
+
 import com.novoda.noplayer.internal.exoplayer.NoPlayerAdsLoader;
 import com.novoda.noplayer.internal.utils.Optional;
 import com.novoda.noplayer.model.AudioTracks;
 import com.novoda.noplayer.model.Bitrate;
+import com.novoda.noplayer.model.Dimension;
 import com.novoda.noplayer.model.PlayerAudioTrack;
 import com.novoda.noplayer.model.PlayerSubtitleTrack;
 import com.novoda.noplayer.model.PlayerVideoTrack;
@@ -14,8 +17,6 @@ import com.novoda.noplayer.model.Timeout;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
-
-import androidx.annotation.FloatRange;
 
 // There are a lot of features for playing and monitoring video.
 @SuppressWarnings("PMD.ExcessivePublicCount")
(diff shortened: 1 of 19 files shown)


core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelectorCreator.java
```diff
@@ -25,8 +26,10 @@ CompositeTrackSelector create(Options options, DefaultBandwidthMeter bandwidthMe
                 Clock.DEFAULT
         );
         DefaultTrackSelector trackSelector = new DefaultTrackSelector(adaptiveTrackSelectionFactory);
+        Dimension maxVideoSize = options.maxVideoSize();
         DefaultTrackSelector.Parameters trackSelectorParameters = trackSelector.buildUponParameters()
                 .setMaxVideoBitrate(options.maxVideoBitrate())
+                .setMaxVideoSize(maxVideoSize.width(), maxVideoSize.height())
```
JozefCeluch: What is the behavior now when we set both max bitrate and max size? Should we add a description of the behavior in the API documentation?

core/src/main/java/com/novoda/noplayer/NoPlayer.java
```diff
@@ -266,6 +267,19 @@
...
+     * those tracks will not be selected.
+     *
+     * @param maxVideoSize The maxium video size in pixels
+     */
+    void setMaxVideoSize(Dimension maxVideoSize);
+
+    /**
+     * Clears the maximum video size, if set.
+     */
+    void clearMaxVideoSize();
```
logicalusman: For my knowledge: 
1- max size has to be set when initializing the player or it can be set anytime?
2- Can a max size be cleared while video in being played?

zegnus: Yes to both. You can set an initial max size when we create the player and you can change the max video size while the video is already playing. Those two scenarios are coded in the demo application ;)

core/src/main/java/com/novoda/noplayer/internal/exoplayer/CompositeTrackSelectorCreator.java
zegnus: Mmm both methods have their own Javadoc. The behaviour will be that both limits will be applied.
- No track will be selected if the bitrate of the track is higher than the set in `setMaxVideoBitrate`
- No track will be selected if the video size is higher than the set in `setMaxVideoSize`

core/src/main/java/com/novoda/noplayer/NoPlayer.java
logicalusman: thanks. Fair enough.

PR #281, opened 2019-10-24
Expose resize mode change

JozefCeluch:
## Problem

no-player doesn't allow clients to set resize mode on exoplayer

## Solution

Expose the resize mode on the `PlayerView` interface. In order to make it typed I created an enum that matches the resize modes provided by [`AspectRatioFrameLayout`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/ui/AspectRatioFrameLayout.html)

### Test(s) added 

none, but I updated the demo

### Screenshots

| Before | After |
| ------ | ----- |
| gif/png _before_ |  |

### Paired with 

nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/PlayerView.java
+++ b/core/src/main/java/com/novoda/noplayer/PlayerView.java
@@ -2,6 +2,7 @@ package com.novoda.noplayer;
 
 import android.view.View;
 
+import com.google.android.exoplayer2.ui.AspectRatioFrameLayout;
 import com.novoda.noplayer.model.TextCues;
 
 public interface PlayerView {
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/novoda/noplayer/PlayerView.java
```diff
@@ -1,6 +1,8 @@
 package com.novoda.noplayer;
 
 import android.view.View;
+
+import com.google.android.exoplayer2.ui.AspectRatioFrameLayout;
```
zegnus: I think we need to remove ExoPlayer dependencies from this public interface, you can have a factory or something like that for your enum, that's private cc @Mecharyry

JozefCeluch: this interface is at the same level as `NoPlayerView` that implements it and uses `AspectRatioFrameLayout` internally. To me it seems like using the constants from `AspectRatioFrameLayout` here is the same internal usage

Mecharyry: 🤔 I take your point @JozefCeluch, `NoPlayerView` is poorly named, it's really `ExoPlayerView` 😂 really the mapping should live in the implementation of the interface and not in the `interface` itself. That way clients that implement the `PlayerView` are free to change the `AspectRatioFrameLayout` for whatever they like.

JozefCeluch: @Mecharyry @zegnus what do you think about this https://github.com/novoda/no-player/pull/281/commits/c0a40a2cf26d6c9c72a20c77847df427ee7e2829?

PR #282, opened 2019-11-04
Metadata callback

laithnurie:
## Problem

Metadata object (Exoplayer) was being passed as a string to the info listener, you can't access data properly within it (to use for analytics reporting etc...).

## Solution

Create NoPlayer Metadata object and parse Exoplayer metadata object to the NoPlayer one.
At some point we probably need to do the same thing for the android media player

### Test(s) added 

Unit tests for the Metadata parser

### Paired with 

Nobody

### DIFF
--- a/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerMetadataListener.java
+++ b/core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerMetadataListener.java
@@ -1,0 +1,27 @@
+package com.novoda.noplayer.internal.exoplayer.forwarder;
+
+import com.google.android.exoplayer2.metadata.Metadata;
+import com.google.android.exoplayer2.metadata.MetadataOutput;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class ExoPlayerMetadataListener implements MetadataOutput {
+
+    private final List<MetadataOutput> listeners = new CopyOnWriteArrayList<>();
+
+    public void add(MetadataOutput listener) {
+        listeners.add(listener);
+    }
+
+    public void remove(MetadataOutput listener) {
+        listeners.remove(listener);
+    }
+
+    @Override
+    public void onMetadata(Metadata metadata) {
+        for(MetadataOutput listener : listeners) {
+            listener.onMetadata(metadata);
+        }
+    }
+}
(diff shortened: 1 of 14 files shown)


hal9002:
Can one of the admins review this PR?

core/src/main/java/com/novoda/noplayer/internal/exoplayer/forwarder/ExoPlayerMetadataListener.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.android.exoplayer2.metadata.Metadata;
+import com.google.android.exoplayer2.metadata.MetadataOutput;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class ExoPlayerMetadataListener implements MetadataOutput {
+
+    private final List<MetadataOutput> listeners = new CopyOnWriteArrayList<>();
```
zegnus: 👍

core/src/main/java/com/novoda/noplayer/metadata/Metadata.java
```diff
@@ -0,0 +1,56 @@
+package com.novoda.noplayer.metadata;
+
+import java.util.List;
+
+public class Metadata {
+
+    private final List<Entry> entries;
```
zegnus: cannot we have two entries, one for `BinaryFrame` and another one for `EventMessage`? I'm asking because from the client point of view you are going to force them to inspect the `Entry` interface in order to find out the to what they need to cast to

danybony: Good point 👍

core/src/main/java/com/novoda/noplayer/metadata/Metadata.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public Metadata(List<Entry> entries) {
+        this.entries = entries;
+    }
+
+    /**
+     * Return a list of entries that are instance of {@link BinaryFrame} or {@link EventMessage}
```
zegnus: 👍

core/src/main/java/com/novoda/noplayer/metadata/Metadata.java
zegnus: Talked with @laithnurie offline. Looks like the client needs a list of events **inorder**, so we are going to add some javadoc on the types returned

zegnus:
test this please

zegnus:
test this please

zegnus:
test this please

PR #287, opened 2019-12-02
Add GitHub Action as a pull request builder

zegnus:
## Problem

We want to move away from local Jenkins to the cloud CI on GitHub

## Solution

- Created a Pull Request builder as GitHub Actions
- At the end of the build, it will upload an artifacts.zip file that will contain all static analysis and tests results even if the build fails

### Test(s) added 

Tested in this PR

### Screenshots

No UI changes

### Paired with 

Nobody

### DIFF
--- a/.github/workflows/pull-request-builder.yml
+++ b/.github/workflows/pull-request-builder.yml
@@ -1,0 +1,53 @@
+name: Pull Request Builder
+
+on:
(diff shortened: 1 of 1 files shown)


zegnus:
@Mecharyry yup, but that's another day PR :) this one is just for building pull requests :)

PR #288, opened 2019-12-20
Move ci to GitHub actions

zegnus:
This PR allows:
- Pull request to be build in GitHub Actions
- Snapshot and Release automatic upload when there is a merge into develop and master
- Notify novoda slack #nos channel

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -36,7 +36,7 @@ buildProperties {
     }
     publish {
         def generateVersion = {
-            long first2020 = new Date().parse("dd.MM.yyy", '22.12.2019').getTime()
+            long first2020 = new Date().parse("dd.MM.yyy", '01.01.2020').getTime()
             long today = new Date().getTime()
             int todaySinceFirst2020InSeconds = (today - first2020) / 1000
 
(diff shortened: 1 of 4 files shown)


core/build.gradle
```diff
@@ -24,13 +36,19 @@ buildProperties {
     }
     publish {
         def generateVersion = {
+            long first2020 = new Date().parse("dd.MM.yyy", '22.12.2019').getTime()
```
Mecharyry: You'll need to explain this to me 😂

zegnus: ui, that was a test xD, changing.

So the actual reason is that as the CI builds in a new machine, there is no way to persist the snapshot version without doing a commit, but a commit will trigger another build and we don't want that. So I decided to use a timestamp, but then the number of the default timestamp is huge and I thought that if I can shorten it then it would be a bit easier to identify. Then what I did is to create a timestamp, in seconds, from first Jan 2020. Fixing now

zegnus: https://github.com/novoda/no-player/pull/288/commits/56a821c75672b3a3c402a307ca3f948904471b64

Mecharyry: Not sure if this would be an issue given it is in seconds but couldn't we potentially have the builds out of order if this is built on machines in different timezones?

zegnus: sorted https://github.com/novoda/no-player/pull/288/commits/56a821c75672b3a3c402a307ca3f948904471b64

PR #292, opened 2020-06-24
Avoid reset state when there is an error

zegnus:
## Problem

The AdvertBreakEnd is not called properly after all adverts time out.
The reason is because we receive an AdvertBreakEnd with the information about the advert, then there is an error and we reset the information of the advert. Then at the end, because we have been reseting the advert info, the AdvertBreakEnd is skipped.

## Solution

This is a temporal solution, I might revert this after further testing client side

Avoid reseting the state on advert

### Test(s) added 

This seems to work, but I might revert this change if we do not fully solve the issue on client side

### Screenshots

No changes

### Paired with 

Nobody

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -67,7 +67,7 @@ android {
     lintOptions {
         lintConfig teamPropsFile('static-analysis/lint-config.xml')
         abortOnError true
-        warningsAsErrors true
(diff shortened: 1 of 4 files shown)


zegnus:
I'm merging this for testing purposes

